PR #42, opened 2023-08-02
Standardize tokens

paulgb:
This standardizes `token` as the name for a type that provides information for connecting to a resource, comprising of "server tokens" for connecting from the server and "client tokens" for connecting to a particular document from the client.

The fields each token has in common are also standardized.

Previously, we referred to a `ClientToken` interchangeably as an "AuthDocResponse" and as a "connection key", which is confusing.

It also changes field names to one-word names: `doc_id` -> `doc`; `base_url` and `endpoint` -> `url`. This way they are idiomatic on both the rust side and JS side.

This is a breaking change, so I wanted to do it before launching.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -29,6 +29,9 @@ jobs:
 
     - name: Install dependencies of JS library
       run: npm ci
+    - name: Test deps
+      run: npm ci
+      working-directory: tests
     - name: Run integration tests
       run: npm test
       working-directory: tests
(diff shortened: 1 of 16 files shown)


paulgb:
It turns out the way the workspace is set up doesn't quite work for recursive dependencies. Tests will refer to the latest version of `@y-sweet/sdk` and `@y-sweet/react` in the repo, but `@y-sweet/react` referred to the version of `@y-sweet/sdk` on npm, which broke the build.

The fix for now is to reproduce `ClientToken` in both repos and drop the inter-dependency because I can't be bothered to keep mucking with npm.

pretentious7 (approved):

PR #43, opened 2023-08-02
add wrapper for server

pretentious7:


js-pkg/server/src/get_binary.js
```diff
@@ -0,0 +1,58 @@
...
+  return url
+}
+
+async function downloadFile(url, filename_or_folder = '.') {
+  const zlib = require('zlib')
+  const { Readable } = require('stream')
+  const { finished } = require('stream/promises')
+  const res = await fetch(url)
+  if (res.status === 404) {
+    throw new Error('no such binary')
```
paulgb, line 25: I suggest:

```javascript
if (res.status === 404) {
    throw new Error(`Tried to download ${url} but the file was not found. It may have been removed.`)
} else if (res.status !== 200) {
    throw new Error(`Error downloading ${url}: server returned ${res.status}`)
}
```

js-pkg/server/src/run.js
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+
+const { spawnSync } = require('node:child_process')
+const { installBinary, binaryExists } = require('./get_binary')
+
+async function runBinary() {
+  try {
+    var binpath = binaryExists()
```
paulgb, line 8: I didn't realize JS would allow this? I would have thought a var would be scoped to the `try {}`, so it would be undefined later. But if it works I'm fine with it.

js-pkg/server/src/get_binary.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+  let tmpdir = await fs.mkdtemp('bin')
```
paulgb, line 27: surprisingly, `mkdtemp` alone does not put the file in `/tmp`. To do this in a platform-agnostic way, let's do this:

```js
 mkdtemp(join(tmpdir(), 'y-sweet-'));
```

paulgb (commented):
Works for me on Mac! A couple small changes but otherwise LGTM. Maybe @rolyatmax will want to do a pass of the npm stuff which he knows better than I do.

js-pkg/server/src/run.js
pretentious7: var is function scope, let is block scope

js-pkg/server/src/get_binary.js
pretentious7: since I wasn't using the tmpfile bit anyway ive removed it,

good to know tho

js-pkg/server/src/get_binary.js
pretentious7: :+1:

paulgb:
Another thought -- if the binary already exists in node_modules but the user updates the y-sweet package, it will run the old binary, which is not ideal. I think we should put the binary name in the filename, like `y-sweet-0.0.1`, to avoid this.

rolyatmax:
@pretentious7 @paulgb It seems like we should try to keep the y-sweet version on npm in line with the y-sweet server's release version. So far, this PR has `npx y-sweet@0.0.5` using `v0.0.2` of the server.

Two possible solutions:

1. Always release a new version of the server when the wrapper's version is updated (and vice versa)
2. Have the npx wrapper's version be something like `v0.0.2-a` where `0.0.2` matches the version of the server its wrapping and `-a`, `-b`, -`c` etc are for follow-on updates that only affect the wrapper.

Other ideas?

pretentious7:
> @pretentious7 @paulgb It seems like we should try to keep the y-sweet version on npm in line with the y-sweet server's release version. So far, this PR has `npx y-sweet@0.0.7` using `v0.0.2` of the server.
> 
> Two possible solutions:
> 
>     1. Always release a new version of the server when the wrapper's version is updated (and vice versa)
> 
>     2. Have the npx wrapper's version be something like `v0.0.2-a` where `0.0.2` matches the version of the server its wrapping and `-a`, `-b`, -`c` etc are for follow-on updates that only affect the wrapper.
> 
> 
> Other ideas?

What's the advantage to having the version numbers mirror each other? 

I'd think the easiest approach would be hardcoding the version in the wrapper, and changing it on upgrade (I can add a script for this) Alternatively could pull the most recent release from github.

paulgb:
> What's the advantage to having the version numbers mirror each other?

If someone wants to run a specific version of the server, or know which version they are running, it will save them from having to comb through the source to see.

paulgb:
> Always release a new version of the server when the wrapper's version is updated (and vice versa)

I'm inclined towards this approach. If it becomes annoying we can automate it.

I don't expect the wrapper script itself to change often once it's stable.

rolyatmax:
> I'm inclined towards this approach. If it becomes annoying we can automate it.

Yeah, I think this is the cleanest.

> What's the advantage to having the version numbers mirror each other?

When people run `npx y-sweet@[SPECIFIC VERSION]` they want to run that specific version of the server.

js-pkg/server/src/get_binary.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+    let dirpath = path.join(modpath, 'y-sweet', 'bin')
+    await fs.mkdir(dirpath, { recursive: true })
+    let filepath = path.join(dirpath, `y-sweet-${version}`)
+    let file = await downloadFile(url, filepath)
+    return file
+  }
+}
+
+const version = require('../package.json').version
```
paulgb, line 52: since it's a global, let's call it `VERSION`.

paulgb: and put it at the top below imports

paulgb (commented):
Couple more nits but otherwise looks good.

rolyatmax (approved):
LGTM! Nice work!

paulgb (approved):

PR #45, opened 2023-08-02
make sidebar work on mobile + mobile styles for all demos

rolyatmax:
This gets the sidebar working on mobile. Still need to do style updates for all the demos and the index page.

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -10,6 +10,8 @@
     "format": "prettier --write \"**/*.{ts,tsx,js,jsx,md}\""
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


paulgb:
This looks good! Haven't tried it on mobile yet but it looks good when I make my window narrow.

feliciachang:
small nit: is it possible to have the body span the height of the screen? 
image

rolyatmax:
> small nit: is it possible to have the body span the height of the screen?

@feliciachang yeah, let me see if I can get that working

PR #47, opened 2023-08-02
try to fix types and tests

rolyatmax:
This makes the tests directory an npm workspace and switches to tsup for building our typescript projects.

To test this out:

1. Pull down this branch
2. Run `npm install` in the root or any workspace (will trigger an npm install for all workspaces)
3. Run `npm run dev` in `js-pkg/react` and/or `js-pkg/sdk`
4. Make some breaking changes
5. Run `npm test` from `tests`

Tests should fail.


paulgb:
I was able to reproduce the issue I hit into before:

- Change `ClientToken` in the sdk to `ClientTokena`
- Change the associated import on the react side
- Bump the version of the sdk to 0.0.3

I suspect that what happened before was that someone bumped the package version but not the dependency. I guess it's partly my expectations being set by Rust, which will prefer the local workspace package if it is available regardless of the version specified (but you have to be explicit about depending on the workspace version). I wonder if there is a way to get similar behavior from npm?

rolyatmax:
> I was able to reproduce the issue I hit into before:
> 
>     * Change `ClientToken` in the sdk to `ClientTokena`
>     * Change the associated import on the react side
>     * Bump the version of the sdk to 0.0.3

Hmm, I did those steps and typechecking and tests pass for me. Maybe we can walk through it together tomorrow. I'm probably not replicating what you did exactly.

> I suspect that what happened before was that someone bumped the package version but not the dependency. I guess it's partly my expectations being set by Rust, which will prefer the local workspace package if it is available regardless of the version specified (but you have to be explicit about depending on the workspace version). I wonder if there is a way to get similar behavior from npm?

When I bump _only_ the sdk's package.json version to 0.0.3 and leave the sdk version in the tests' package.json at 0.0.2, then npm install, I find that the tests' sdk points to the local sdk package - version 0.0.3. Is that the behavior you're wanting?

paulgb:
Here's an example repro: https://www.loom.com/share/c4caf8b1f4334063a2231fb6af4c1399?sid=44c91cd7-62bf-41a9-be41-21aeae4ea796

Not shown, but I run `rm -rf **/node_modules **/dist` at the beginning. On commit 382c9224.

rolyatmax:
> Here's an example repro: https://www.loom.com/share/c4caf8b1f4334063a2231fb6af4c1399?sid=44c91cd7-62bf-41a9-be41-21aeae4ea796

Ahh gotcha - I thought you were talking about a problem you ran into running tests. So two things:

1. That `examples/package-lock.json` file shouldn't be there. I meant to remove that on a previous PR. Pushed a commit for that just now.
2. When you update the version of a local package, you have to update that package's version wherever it appears as a dependency in your other packages. If you don't, it'll install the old version from npm and use that. (I guess this isn't how Rust workspaces work? It is how _I_ would expect things to work, though. Otherwise, you can ship packages that point to old dependencies - the combination would work locally because you would be using a different version than what's actually in your package.json. It would then be broken for all your users who are getting the version defined in the package json.)

paulgb:
> I guess this isn't how Rust workspaces work? I think it's how I would expect things to work, though. Otherwise, you can ship packages that point to old dependencies - the combination would work locally because you would be using a different version than what's actually in your package.json. It would then be broken for all your users who are getting the version defined in the package json.

In rust the way it works is that a dependency can refer to both a version and a path [(like this)](https://github.com/drifting-in-space/y-sweet/blob/f0ca78141ecd7024bdaea422b7db6ad5ef475441/crates/y-sweet/Cargo.toml#L29). The version pointed to by `path` takes precedence _except_ when publishing to crates.io. When you push to crates.io, it ignores `path` and downloads the `version` of the named crate from crates.io.

It ends up working out pretty well in practice, because:
- The rust ecosystem generally embraces semver
- Unless you pin a version explicitly, the cargo will take the most recent `patch` version for the `major`/`minor` pair, and will update to the latest patch again whenever you run `cargo update`
- Cargo builds packages when it publishes them, so if there is a semver violation in an interface (rather than just a behavior) it will be caught there
- `cargo-workspaces` can automatically build a DAG and push packages in the right order

In the case I ran into it, I didn't touch `package.json` at all -- when it was checked in in #20, `@y-sweet/react` [depended on](https://github.com/drifting-in-space/y-sweet/blob/bf330f951664321e8a4580674fcde0208d38b473/js-pkg/react/package.json#L42) `@y-sweet/sdk` 0.0.1, but the version of `@y-sweet/sdk` in the same commit is [0.0.2](https://github.com/drifting-in-space/y-sweet/blob/bf330f951664321e8a4580674fcde0208d38b473/js-pkg/sdk/package.json#L9)

Unless something already exists, we could write a check of our own that:
- scans the workspace for package.json files
- collects the name and version of each package
- loops over dependencies of each package
- if it recognizes a package as belonging to this repo
- fails if any package doesn't match

I might try creating this, actually.

paulgb:
Ok, I had a go in #51. One thing I'm unsure of is whether npm will do the right thing if we specify "^0.0.2" instead of "0.0.2" but the local package is "0.0.3" -- so currently, the test goes for strict equality, but it can probably be relaxed (which is nice because it won't bug us about patch releases)

paulgb (approved):

PR #49, opened 2023-08-02
Rename crates

paulgb:


.github/workflows/release.yml
```diff
@@ -102,22 +102,22 @@ jobs:
...
+        mv macos-binary/y-sweet.gz y-sweet-macos-x64.gz
     - name: Release
       uses: softprops/action-gh-release@v1
       with:
         name: ${{ github.event.inputs.version }}
         draft: true
         files: |
-            y-sweet-server-win-x64.exe.gz
-            y-sweet-server-linux-x64.gz
-            y-sweet-server-macos-x64.gz
```
rolyatmax, line 123: Do we need to update the npx wrapper? https://github.com/drifting-in-space/y-sweet/blob/main/js-pkg/server/src/get_binary.js#L16

rolyatmax (approved):
LGTM aside from one comment there about updating the npx wrapper to download the renamed binaries.

.github/workflows/release.yml
paulgb: good catch.

PR #55, opened 2023-08-03
demo copy link and one line descriptions

feliciachang:


examples/src/app/page.tsx
```diff
@@ -10,10 +10,26 @@ export default async function Home({ searchParams }: HomeProps) {
...
-        <ProjectLink name="Code Editor" url="/code-editor" />
+        <ProjectLink
+          name="Color Grid"
+          url="/color-grid"
+          description="Drop different shades of pink into a grid."
+        />
+        <ProjectLink
+          name="To Do List"
+          url="/to-do-list"
+          description="Write and clear a to do list."
```
rolyatmax, line 21: Alternative wording?: "Create and edit items in a to do list"?

examples/src/app/page.tsx
```diff
@@ -10,10 +10,26 @@ export default async function Home({ searchParams }: HomeProps) {
...
+        />
+        <ProjectLink
+          name="Text Editor"
+          url="/text-editor"
+          description="A collaborative text editor that uses a quill yjs binding."
```
rolyatmax, line 26: What do you think about: "A collaborative text editor built on top of the open-source Quill package." ?

examples/src/app/page.tsx
```diff
@@ -10,10 +10,26 @@ export default async function Home({ searchParams }: HomeProps) {
...
+        />
+        <ProjectLink
+          name="Code Editor"
+          url="/code-editor"
+          description="A collaborative code editor that uses a codemirror yjs binding."
```
rolyatmax, line 31: And maybe "A collaborative code editor built on top of the open source CodeMirror package." ?

examples/src/components/CopyLink.tsx
```diff
@@ -0,0 +1,33 @@
+'use client'
+import { useState } from 'react'
+
+export default function CopyLink() {
+  const [buttonState, setButtonState] = useState('Copy link')
```
paulgb, line 5: nit: rather than setting UI text in state, I would prefer here to do something like:

```tsx
const [copied, setCopied] = useState(false)
```

and then when rendering the button

```tsx
{clicked ? 'Copy link' : 'Copied!'}
```

In this case the component is simple enough that it doesn't make a big difference, but this decouples the state from the copy which makes some things easier, like if we wanted to translate the UI down the road.

examples/src/components/CopyLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <span className="text-pink-900 pb-2">
+        Open this doc in a new window or share with a friend:
+      </span>
+      <button
+        className="w-28 py-2 rounded-lg bg-pink-950 text-white transition-all hover:bg-pink-900"
+        onClick={copyLinkToClipboard}
+      >
+        {buttonState}
+      </button>
+    </div>
```
rolyatmax, line 31: I like this idea a lot, but I kinda feel like it visually competes with the demos themselves too much:

Screenshot 2023-08-04 at 11 19 21 AM

I played around with making it smaller like this:

Screenshot 2023-08-04 at 11 21 52 AM

I also tried moving it up next to the title, but it still didn't look right to me.

I like what Liveblocks does by placing it in their top bar:

liveblocks-share

Any thoughts?

feliciachang: i like these approaches. i can work on the simpler inline version for now to get this out sooner (and follow up with something more sophisticated if needed)

rolyatmax (approved):

PR #56, opened 2023-08-05
Use URL instead of JSON string for config

paulgb:
When writing docs, I found that configuring the server by passing JSON as a string is confusing. Instead, we can store the server token directly in the URL and extract it.

So instead of:

```bash
Y_SWEET_CONFIG='{"endpoint":"http://127.0.0.1:8080/","token":"AAAgeGUDfVahsev4pDpmv39ybuWJHyAdrPKjSfEA8m4zTRI="}' npm run dev
```

we now say:

```bash
Y_SWEET_CONFIG=http://AAAgEaepIDE1UEQ17qNreTVPdYD5owf%2F0Fzq7RGN5fekvDg%3D@127.0.0.1:8080/ npm run dev
```

This PR also cleans up a few things about config:

- `USE_HTTPS` is no longer required on the Cloudflare workers side; instead I found a CF-specific header that we can infer it from.
- On the native side, `--use-https` is replaced with `--url-prefix`. Using HTTPS implies that we are behind a reverse proxy, which also means we might have a URL prefix, so that's now supported.
- The server token is now printed when the server starts (unless `--prod` is passed) rather than when the auth token is generated.


crates/y-sweet/src/server.rs
```diff
@@ -265,8 +266,16 @@ async fn auth_doc(
...
-    let schema = if server_state.use_https { "wss" } else { "ws" };
-    let url = format!("{schema}://{host}/doc/ws");
+    let url = if let Some(url_prefix) = &server_state.url_prefix {
+        let mut url = url_prefix.clone();
+        let scheme = if url.scheme() == "https" { "wss" } else { "ws" };
+        url.set_scheme(scheme).unwrap();
+        url.join("/doc/ws").unwrap().to_string();
+        url.to_string()
+    } else {
+        format!("ws://{host}/doc/ws")
```
rolyatmax, line 276: This `url` is used by the client, right? If so, will running the server in prod essentially mean always passing a `url_prefix`? Assuming that running in prod will always mean using HTTPS.

paulgb: Yes. Maybe if `--prod` is passed without a `--url-prefix`, we should print a warning or even fail to start the server?

rolyatmax: Maybe it makes sense to disentangle the two options and have the url prefix just be a path prefix? And then HTTPS is on by default unless you pass `--http` or something like that?

rolyatmax: Yeah, true

rolyatmax (approved):

PR #59, opened 2023-08-05
Allow region to be specified via env vars

paulgb:
- Allow region to be specified via env vars.
- Remove some panics and add some error context.
- Fixed an issue where 404s were not being propagated correctly from s3-rust because a feature flag was disabled.

### DIFF
--- a/crates/y-sweet-core/src/sync_kv.rs
+++ b/crates/y-sweet-core/src/sync_kv.rs
@@ -1,5 +1,5 @@
 use crate::store::Store;
-use anyhow::Result;
+use anyhow::{Context, Result};
 use std::{
     collections::BTreeMap,
     convert::Infallible,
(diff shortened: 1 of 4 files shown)


crates/y-sweet-core/src/sync_kv.rs
```diff
@@ -26,12 +26,13 @@ impl SyncKv {
...
-        let data = if let Some(snapshot) = store.get(&key).await? {
-            tracing::info!(size=?snapshot.len(), "Loaded snapshot");
-            bincode::deserialize(&snapshot)?
-        } else {
-            BTreeMap::new()
-        };
+        let data =
+            if let Some(snapshot) = store.get(&key).await.context("Failed to get from store.")? {
+                tracing::info!(size=?snapshot.len(), "Loaded snapshot");
+                bincode::deserialize(&snapshot).context("Failed to deserialize.")?
```
rolyatmax, line 32: Might be worth logging more context here and above so we can look into which key didn't exist or which key's value failed to deserialize?

crates/y-sweet/src/main.rs
```diff
@@ -51,12 +51,18 @@ enum ServSubcommand {
...
     if store_path.starts_with("s3://") {
-        let region = Region::UsEast1;
-
+        let region = match Region::from_default_env() {
+            Ok(region) => {
+                tracing::info!(region=?region, "Using region from environment.");
+                region
+            }
+            Err(e) => {
+                tracing::warn!(error=?e, "Failed to get region from environment, using default.");
```
rolyatmax, line 60: Worth mentioning here that the default value is us-east-1?

rolyatmax (approved):

crates/y-sweet/src/main.rs
paulgb: Done.

crates/y-sweet-core/src/sync_kv.rs
rolyatmax: Gotcha - does that include the key or the value?

PR #63, opened 2023-08-07
Initial tree CRDT demo

paulgb:


examples/src/app/tree-crdt/util.ts
```diff
@@ -0,0 +1,9 @@
+export function randomString() {
+  return Math.random().toString(36).substring(2, 7)
+}
+
+export function randomColor() {
+  const hue = Math.random() * 360
+  const value = Math.random() * 0.5 + 0.5
+  return `hsl(${hue}, 100%, ${value * 100}%)`
```
rolyatmax, line 8: Nit: I think you'll get better colors with a saturation and lightness each between 50-75%.

examples/src/app/tree-crdt/TreeView.tsx
```diff
@@ -0,0 +1,222 @@
...
+    },
+  }
+}
+
+function Arrow(props: { origin: Point; target: Point; offset?: number; animate: boolean }) {
+  let length = Math.sqrt(
+    (props.target.x - props.origin.x) ** 2 + (props.target.y - props.origin.y) ** 2,
+  )
+
+  if (length < (props.target.radius || 0)) {
```
rolyatmax, line 41: When would this be true? How could the distance between two nodes be less than the radius of one of the nodes (unless the layout was messed up) or less than 0?

examples/src/app/tree-crdt/TreeView.tsx
```diff
@@ -0,0 +1,222 @@
...
+export function TreeView() {
+  const [tree, setTree] = useState<YTreeNode | null>(null)
+  const [_, bumpTreeVersion] = useState(0)
+  const treeMap = useMap<Y.Map<any>>('tree')
+
+  useEffect(() => {
+    let tree = new YTree(treeMap)
+    tree.setOnChange(() => {
+      bumpTreeVersion((v) => v + 1)
+    })
```
rolyatmax, line 202: Do you need to unsubscribe this `onChange` callback on unmount?

examples/src/app/tree-crdt/TreeView.tsx
paulgb: If you create enough children of one node, there can be less than the radius between them. Maybe that counts as “layout is messed up”, but it's easy enough to solve for.

image

0 is used here because the radius of a point with no radius set is 0. length will never be less than 0, but we still want the expression to evaluate to `false` rather than `undefined`. We use a zero radius point when drawing the arrow to the mouse position instead of locking it on to a circle.

examples/src/app/tree-crdt/tree_layout.ts
rolyatmax, line 1: ooo this is cool

examples/src/app/tree-crdt/radixpq.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (this.subIter) {
+        const result = this.subIter.next()
+        if (!result.done) {
+          return { value: result.value[1], done: false }
+        }
+        this.subIter = null
+      } else {
+        const result = this.iter.next()
+        if (result.done) {
+          return { value: undefined!, done: true }
```
rolyatmax, line 48: Maybe a better return type for this function would be?:

```ts
next(): { value: T; done: false } | { value: void; done: true }
```

examples/src/app/tree-crdt/TreeView.tsx
rolyatmax: Ahh gotcha

examples/src/app/tree-crdt/ytree.ts
rolyatmax, line 1: nice!

rolyatmax (approved):
Aside from a few unaddressed comments, this LGTM!

examples/src/app/tree-crdt/radixpq.ts
paulgb: Good idea.

examples/src/app/tree-crdt/TreeView.tsx
paulgb: I think it's probably ok not to here because the lifetime of `bumpTreeVersion` and `tree` should always be the same as the life of the component, but it doesn't hurt to so I've done it.

PR #65, opened 2023-08-11
Voxel demo

paulgb:
Demo: https://y-sweet-demos-git-paulgb-voxel-demo-drifting-corp.vercel.app/voxels


examples/package.json
```diff
@@ -12,17 +12,24 @@
...
+    "@react-three/fiber": "^8.13.6",
+    "@types/react-color": "^3.0.6",
+    "@uiw/react-color": "^1.3.3",
     "@y-sweet/react": "0.0.4",
     "@y-sweet/sdk": "0.0.4",
     "autoprefixer": "10.4.14",
     "codemirror": "^5.65.14",
+    "i": "^0.3.7",
     "next": "13.4.10",
+    "npm": "^9.8.1",
```
rolyatmax, line 25: Are `npm` and `i` supposed to be here?

paulgb: uh weird, I must have somehow run `npm i npm i`?? Good catch.

js-pkg/react/src/main.tsx
```diff
@@ -31,18 +31,25 @@ export function useAwareness(): Awareness {
...
 
-export function usePresence<T extends Record<string, any>>(): [
-  Map<number, T>,
-  (presence: T) => void,
-] {
+type UsePresenceOptions = {
+  excludeSelf?: boolean
+}
+
+export function usePresence<T extends Record<string, any>>(
```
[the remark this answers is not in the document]
rolyatmax: Why do you think so?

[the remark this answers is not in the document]
paulgb: I tend to agree actually, opened #66.

examples/package.json
```diff
@@ -12,6 +12,10 @@
   "dependencies": {
     "@headlessui/react": "^1.7.16",
     "@heroicons/react": "^2.0.18",
+    "@react-three/drei": "^9.80.2",
+    "@react-three/fiber": "^8.13.6",
+    "@types/react-color": "^3.0.6",
+    "@uiw/react-color": "^1.3.3",
```
rolyatmax, line 18: Why not use `react-color`? 

I don't know if this is related, but I noticed the color picker is off:
Screenshot 2023-08-11 at 10 07 33 AM

js-pkg/react/src/main.tsx
```diff
@@ -31,18 +31,25 @@ export function useAwareness(): Awareness {
...
+  options?: UsePresenceOptions,
+): [Map<number, T>, (presence: T) => void] {
   const awareness = useAwareness()
   const [presence, setPresence] = useState<Map<number, T>>(new Map())
 
+  const excludeSelf = options?.excludeSelf ?? true
```
rolyatmax, line 44: Minor, but if the default is to exclude self, maybe this option could be called `includeSelf`?

js-pkg/react/src/main.tsx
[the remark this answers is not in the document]
pretentious7: > Why do you think so?

So it's easier to revert in isolation, since we squash commits.

rolyatmax (approved):
A few comments there, otherwise LGTM

js-pkg/react/src/main.tsx
paulgb: good call, I like the principle that bools should be oriented such that the default is false.

examples/package.json
paulgb: `react-color` hasn't been touched in two years and causes noisy warnings with recent versions of react ([github issue](https://github.com/casesandberg/react-color/issues/888))

paulgb: oh, I see what you mean, the colors don't line up by column

PR #66, opened 2023-08-11
Only show presence of other users by default

paulgb:
In most cases, the presence indicator for the local user does not follow the same code path as remote users (text cursors, pointers, etc.), so this removes the local presence by default.

### DIFF
--- a/js-pkg/react/src/main.tsx
+++ b/js-pkg/react/src/main.tsx
@@ -31,18 +31,25 @@ export function useAwareness(): Awareness {
   return yjsCtx.provider.awareness
 }
 
-export function usePresence<T extends Record<string, any>>(): [
-  Map<number, T>,
-  (presence: T) => void,
-] {
+type UsePresenceOptions = {
+  excludeSelf?: boolean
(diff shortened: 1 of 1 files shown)


js-pkg/react/src/main.tsx
```diff
@@ -31,18 +31,25 @@ export function useAwareness(): Awareness {
...
+  excludeSelf?: boolean
+}
+
+export function usePresence<T extends Record<string, any>>(
+  options?: UsePresenceOptions,
+): [Map<number, T>, (presence: T) => void] {
   const awareness = useAwareness()
   const [presence, setPresence] = useState<Map<number, T>>(new Map())
 
+  const excludeSelf = options?.excludeSelf ?? true
```
rolyatmax, line 44: Same comment as [here](https://github.com/drifting-in-space/y-sweet/pull/65#discussion_r1291393193)

rolyatmax (approved):
One comment there. LGTM, though we should make sure the js packages eventually get bumped and published.

PR #67, opened 2023-08-12
Allow obtaining a setPresence handle without subscribing to presence

paulgb:
Introduces `useGetPresence` which gets and subscribes to presence, and `usePresenceSetter` which returns a setter but does not subscribe. This allows the presence rerender to be pushed down the component tree while using the setter elsewhere in the tree. 

The existing `usePresence` hook behavior is unchanged, but it is rewritten in terms of the new hooks. I considered deprecating it, but since `useGetPresence` and `usePresenceSetter` are both generic functions, I think it's useful to have a way to call them both with the same type parameter -- open to thoughts on this.

### DIFF
--- a/examples/src/app/presence/Presence.tsx
+++ b/examples/src/app/presence/Presence.tsx
@@ -9,13 +9,16 @@ export function Presence() {
   const myColor = useRef(COLORS[Math.floor(Math.random() * COLORS.length)])
   const [presence, setPresence] = usePresence<{ x: number; y: number; color: string }>()
 
-  const updatePresence = useCallback((e: React.MouseEvent<HTMLDivElement>) => {
-    setPresence({
-      x: e.clientX - e.currentTarget.offsetLeft,
-      y: e.clientY - e.currentTarget.offsetTop,
-      color: myColor.current,
-    })
-  }, [setPresence])
+  const updatePresence = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      setPresence({
+        x: e.clientX - e.currentTarget.offsetLeft,
+        y: e.clientY - e.currentTarget.offsetTop,
+        color: myColor.current,
+      })
+    },
+    [setPresence],
+  )
 
   return (
     <div
--- a/examples/src/app/presence/Presence.tsx
+++ b/examples/src/app/presence/Presence.tsx
@@ -15,7 +15,7 @@ export function Presence() {
       y: e.clientY - e.currentTarget.offsetTop,
       color: myColor.current,
     })
-  }, [])
+  }, [setPresence])
 
   return (
     <div
--- a/examples/src/app/presence/Presence.tsx
+++ b/examples/src/app/presence/Presence.tsx
@@ -1,13 +1,16 @@
 'use client'
 
-import { usePresence } from '@y-sweet/react'
+import { usePresence, usePresenceSetter } from '@y-sweet/react'
 import { useCallback, useRef } from 'react'
 
 const COLORS = ['bg-red-500', 'bg-blue-500', 'bg-green-500', 'bg-purple-500', 'bg-pink-500']
 
+type Presence = { x: number; y: number; color: string }
+
 export function Presence() {
   const myColor = useRef(COLORS[Math.floor(Math.random() * COLORS.length)])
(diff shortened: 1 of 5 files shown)


examples/src/app/presence/Presence.tsx
```diff
@@ -9,13 +9,16 @@ export function Presence() {
...
-  }, [])
+  const updatePresence = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      setPresence({
+        x: e.clientX - e.currentTarget.offsetLeft,
+        y: e.clientY - e.currentTarget.offsetTop,
+        color: myColor.current,
+      })
+    },
+    [setPresence],
```
rolyatmax, line 20: Minor: can remove `setPresence` from deps since state setters are guaranteed by React to have consistent references across renders. https://react.dev/learn/synchronizing-with-effects#why-was-the-ref-omitted-from-the-dependency-array

js-pkg/react/src/main.tsx
```diff
@@ -64,16 +79,18 @@ export function usePresence<T extends Record<string, any>>(
...
-    [awareness],
-  )
+  return presence
+}
+
+export function usePresence<T extends Record<string, any>>(): [
+  Map<number, T>,
+  (presence: T) => void,
+] {
+  const awareness = useAwareness()
```
rolyatmax, line 89: Looks like this line can be dropped?

examples/src/app/presence/Presence.tsx
rolyatmax: Oh wait - duh - this is a setter from _our_ hook. Maybe we should make sure this stays referentially stable? 🤔 I'm leaving this in for now, though.

js-pkg/react/src/main.tsx
```diff
@@ -64,16 +79,18 @@ export function usePresence<T extends Record<string, any>>(
...
-      if (awareness) {
-        awareness.setLocalState(localState)
-      }
-    },
-    [awareness],
-  )
+  return presence
+}
+
+export function usePresence<T extends Record<string, any>>(): [
```
rolyatmax, line 85: Minor: I kinda think it might be better to not have a `usePresence` hook at all. In the best case, devs can write one line instead of two:

```ts
const presence = useGetPresence()
const setPresence = usePresenceSetter()
```

In the worst case, they are subscribing to presence updates unnecessarily and their app slows way down and they don't know why, so they just blame it on y-sweet.

I don't think the upside is enough to offset the downside.

js-pkg/react/src/main.tsx
```diff
@@ -35,9 +35,24 @@ type UsePresenceOptions = {
...
+        awareness.setLocalState(localState)
+      }
+    },
+    [awareness],
+  )
+
+  return setLocalPresence
+}
+
+export function useGetPresence<T extends Record<string, any>>(
```
rolyatmax, line 53: If we do get rid of the `usePresence` hook like I suggested above, I think we should rename this to `usePresence`. To me `useThing` means "give me `Thing`". So `useGetThing` reads as "give me a function that gets `Thing`".

Just my two cents.

examples/src/app/voxels/VoxelEditor.tsx
rolyatmax, line 1: nice!

rolyatmax (approved):
A few thoughts there - let me know what you think

examples/src/app/presence/Presence.tsx
paulgb: Yeah, if it's not there we get a linter warning. I wonder if there is an annotation or something we can use to tell the linter that it's referentially stable?

js-pkg/react/src/main.tsx
paulgb: My only hesitation on this is that I think that discourages people from specifying the generic type, since they have to keep it in sync between two locations. But that's a minor thing, they can always create a type alias.

paulgb:
@rolyatmax I implemented your suggestions, thanks!

examples/src/app/presence/Presence.tsx
rolyatmax: Yeah, the linter might not be able to work that out. But also our `setPresence` isn't technically referentially stable since it changes whenever the awareness context changes.

PR #68, opened 2023-08-12
fix: get_binary.js download url

knownasilya:
The releases urls don't match


paulgb:
Thanks for reporting, the version here matched a change to the release script but we had not done a release since the change. We have made a release to fix the issue, if you `git pull` on main it should work now.

PR #69, opened 2023-08-12
fix: client token properties

knownasilya:


paulgb:
Hi @knownasilya, thanks for the PRs and sorry you ran into these issues. It looks like they arose because of version differences between the git repo and the packages deployed on npm and server binaries. We generally try to avoid breaking changes between the repo and deployed code, but we broke things here.

Everything should be up to date now, just make sure to pull the latest from the repo and if you have downloaded any packages that they are on the latest version (0.0.5 for `@y-sweet/*` and 0.0.9 for `y-sweet` [server])

PR #72, opened 2023-08-14
Abhi/release script

pretentious7:


paulgb:
As discussed offline, I don't want to introduce a shell script into the repo and this is not high priority.

PR #74, opened 2023-08-15
Fix url prefix bug

paulgb:
This fixes a bug where the path component of a configured `url_prefix` was not preserved.

Also adds some error logging and metadata around configuration errors.

### DIFF
--- a/crates/y-sweet-worker/src/lib.rs
+++ b/crates/y-sweet-worker/src/lib.rs
@@ -130,10 +130,12 @@ async fn auth_doc(
         match parsed.scheme() {
             "http" => parsed.set_scheme("ws").map_err(|_| Error::InternalError)?,
             "https" => parsed.set_scheme("wss").map_err(|_| Error::InternalError)?,
-            _ => return Err(Error::ConfigurationError {
-                field: "url_prefix".to_string(),
-                value: url_prefix.to_string(),
-            }),
+            _ => {
+                return Err(Error::ConfigurationError {
+                    field: "url_prefix".to_string(),
+                    value: url_prefix.to_string(),
+                })
+            }
         };
         format!("{parsed}/doc/ws")
     } else {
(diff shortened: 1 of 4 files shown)


crates/y-sweet-worker/src/lib.rs
```diff
@@ -123,15 +123,21 @@ async fn auth_doc(
...
             "https" => parsed.set_scheme("wss").map_err(|_| Error::InternalError)?,
-            _ => return Err(Error::ConfigurationError),
+            _ => {
+                return Err(Error::ConfigurationError {
+                    field: "url_prefix".to_string(),
+                    value: url_prefix.to_string(),
+                })
+            }
         };
-        let result = parsed.join("/doc/ws").unwrap();
```
pretentious7, line 132: what is the bug here exactly?

crates/y-sweet-worker/src/lib.rs
```diff
@@ -123,15 +123,21 @@ async fn auth_doc(
...
-
-        result.to_string()
+        format!("{parsed}/doc/ws")
```
pretentious7, line 140: doesn't this break trailing slashes?
if url_prefix is "https://example.com/a/b/",
this will give "https://example.com/a/b//doc/ws", right?

crates/y-sweet-worker/src/lib.rs
paulgb: `Url::join` works differently than a filesystem path join; if the string being joined starts with a `/`, it replaces the whole path. It aims to be consistent with how a browser would interpret a relative link from a website, in which case a leading `/` means to interpret it as relative to the _domain_ (technically, the authority); not the current path.

We could get the correct behavior here by:
- omitting `/` from the beginning of the joined string
- adding a `/` to the end of the URL (so that the project ID did not get replaced by the new path)

But I think it's clearer and less fragile just to use string manipulation.

pretentious7:
This lgtm, but I think we should have a config error if the url_prefix passed in ends with a /

paulgb:
> This lgtm, but I think we should have a config error if the url_prefix passed in ends with a /

Currently, `url_prefix` can't be configured through the environment; this is only exposed to programs that use this as a library.

pretentious7 (approved):

PR #75, opened 2023-08-15
Improve experience when examples error

paulgb:
Currently if you try to run examples and it can't find the y-sweet server, you get an ugly nextjs error. This catches that error in the examples and shows a message like this:

image

We may want to generalize this into the `@y-sweet/sdk` code, but it's a bit more delicate because:
- We can't assume styles there
- Production users may not want to reveal as much as we do here
- I think (?) the next.js version of fetch is different than the node / deno / bun versions when it comes to the error returned. This works with the next.js fetch, but I'm not sure if it generalizes. In theory, the sdk should not contain next.js-specific code.


rolyatmax:
I like this idea a lot! I was thinking about another approach that might work which would (1) allow us to leave the examples untouched, (2) provide better error messages to everyone using the SDK library, and (3) make use of Nextjs error boundaries:

What if we catch the fetch error in the SDK functions, then use logic you've written to derive a better error message and throw _that_ error. Then, in our `app` directory, we just need to add an `error.tsx` file that looks like:

```tsx
'use client'

export default function Error({ error }: { error: Error }) {
  return (
    <div className="p-10">
      <div className="rounded-md bg-red-50 px-6 py-4 text-md text-red-700">
        <h3 className="font-medium text-red-800 mb-2">
          Error connecting to y-sweet server
        </h3>
        <p>{error.message}</p>
      </div>
    </div>
  )
}
```

rolyatmax:
> I would prefer this approach, but I would like the SDK to work with both node.js and next.js and potentially other runtimes like Deno. As far as I can tell, the fetch itself is compatible with all of these, but next.js has its own (crappy) way of sending errors back that is incompatible with the other approaches.

My understanding is that NextJS only extends the `fetch` functionality. We wouldn't be making use of any of those NextJS-only pieces.

Just to be clear, I'm suggesting we do something like this in the SDK:

```ts
    try {
      var result = await fetch(...)
    } catch (err) {
      const extraContext = getErrorMsg(err, hasToken)
      throw new Error(`${err?.message}\n${extraContext}`)
    }
    if (!result.ok) {
      throw new Error(`Failed to fetch ${url}: ${result.status} ${result.statusText}`)
    }
```

where `getErrorMsg` looks like this:

```ts
function getErrorMsg(error: Error, hasToken: boolean) {
  if (error.cause?.code === 'ECONNREFUSED') {
    let { address, port } = error.cause
    return address === '127.0.0.1'
      ? `It looks like you are trying to connect to a local server, but one isn’t running. Run npx y-sweet serve ${port !== 8080 ? ` --port ${port}` : ''} to run a local server.`
      : `Try curl ${address}:${port} to verify the server is running.`
  }
  if (error.toString().includes('401 Unauthorized')) {
    return hasToken 
    ? 'The server token provided was not accepted by the y-sweet server.'
    : 'The server expects an authorization header, but the request does not include one.'
  }
  return 'Fetch to y-sweet server failed.'
}
```

I'm pretty sure everything there works in NodeJS. If you are NextJS, that error is stripped of some properties (not the `message`, though) and serialized & sent to the client to be rendered into our examples' `error.tsx` component. If folks are using NextJS, they can write their own error components if they'd like, too. If folks aren't using NextJS, then they at least have better error messages when `getOrCreateDoc` throws.

paulgb:
@rolyatmax building on your changes, I moved the `YSweetErrorPayload` parsing into `YSweetErrorPayload` itself, so that the error handler does not have to contain message parsing logic and message parsing code can live next to message serializing code, so that we don't modify one without modifying the other. We could have tests for this to ensure that future PRs don't break a round-trip through `getMessage` / `fromMessage`, but we aren't running `examples/` tests in CI/CD yet so that would be a prereq anyway.

rolyatmax (approved):
I like it!

PR #76, opened 2023-08-15
Allow url-safe base64 for auth

paulgb:
Currently we generate URLs that look like this:

```
http://2Fl8XmOtwEMXkXLIlWwWi7kylAXTuJ+PPyQBWKM%3D@127.0.0.1:8080/project/C8az-GJYXKPNyYP4HRgyl/
```

These are uglier than they need to be because they use non-URL-safe characters. If we instead use URL-safe base64 without padding, we could create a slightly nicer looking URL:

```
http://2Fl8XmOtwEMXkXLIlWwWi7kylAXTuJ-PPyQBWKM@127.0.0.1:8080/project/C8az-GJYXKPNyYP4HRgyl/
```

Merely switching would work, but break existing tokens, so instead I opt to (at least for an interim period) accept both encodings.

In order to accept both encodings for server tokens, instead of comparing the base64-encoded strings, we need to actually validate the token. This has the benefit of enabling expiring server tokens.

I didn't do this initially because I thought HTTP required a Bearer token to be (standard) base64 encoded, but [the RFC](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) merely requires that the token match a regular expression which includes both standard and URL-encoded base64.

### DIFF
--- a/crates/y-sweet-core/src/auth.rs
+++ b/crates/y-sweet-core/src/auth.rs
@@ -1,4 +1,8 @@
-use base64::{engine::general_purpose, Engine};
+use base64::{
+    alphabet::{STANDARD, URL_SAFE},
+    engine::{general_purpose, DecodePaddingMode, GeneralPurpose, GeneralPurposeConfig},
+    Engine,
+};
 use bincode::Options;
 use rand::Rng;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


rolyatmax:
> The rfc had this regex for the token
> 
> > b64token = 1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" ) *"="
> 
> Doesn't that not allow url encoding?

@pretentious7 I think the token itself was always base-64 encoded, but when included in the URL, it was url-encoded which creates an ugly-looking URL. It's url-decoded into the original base-64 when it's read on the other side. I think the goal of this PR is to generate tokens that use only characters in the base-64 set that are _also_ url-safe. That way we comply with the RFC _and_ have pretty URLs.

paulgb:
Yes @rolyatmax has it exactly right.

rolyatmax (commented):
I'm going to leave this one for @pretentious7 to review

pretentious7:
LGTM!

PR #77, opened 2023-08-15
a proposal for changes to #75

rolyatmax:
Here's a proposal for changes to #75 that would allow us to keep the example code minimal and straight-forward (No indirection with `WrappedDocProvider`.)

Another option would be https://github.com/drifting-in-space/y-sweet/commit/c0b5608224a3da2783d31573457e9375f7a6b8f8


examples/src/app/error.tsx
```diff
@@ -0,0 +1,41 @@
+'use client'
+
+function Samp(props: { children: React.ReactNode }) {
+  return <samp className="bg-red-100 p-1 rounded-md font-bold">{props.children}</samp>
+}
+
+export default function Error(props: { error: any }) {
+  let message = <p>Check the server console for more information.</p>
+
+  const errMsg = props.error.message
```
paulgb, line 10: if we are only using this for `startsWith` calls that extract the code, why not just access the code field directly here?

examples/src/app/error.tsx
```diff
@@ -0,0 +1,41 @@
...
+  if (errMsg.startsWith('ServerRefused') && errMsg.includes('127.0.0.1')) {
```
paulgb, line 11: I don't like using string matching this way, especially when we have the data available as structured data inside the `props.error` value, unless I'm missing something?

paulgb:
I didn't know about `error.tsx`, I agree this is cleaner, but I have some questions about why we are parsing data out of the text when it's already available as structured data in the error.

examples/src/app/error.tsx
rolyatmax: The `code` field is not available on the serialized version of the Error that Next sends down to the client, unfortunately.

rolyatmax:
Unfortunately, `message` is the only property that gets serialized and sent to the client that you really have control over. (And even then, only in development.) More info: https://nextjs.org/docs/app/building-your-application/routing/error-handling#handling-server-errors

We could also consider the second option I posted in the description above that generates the message text that ultimately appears in the UI entirely in the SDK Error that's thrown. Personally, I think that error text is perfectly fine for the sdk, especially since it's never user-facing and is quite useful for the developer.

paulgb:
Ah, gotcha. Kind of an unfortunate design decision on the part of next.js -- seems reasonable to share nothing in prod, but in dev it seems pretty arbitrary to assume that the `message` field doesn't have sensitive information.

One thing I like about this approach and the original (as opposed to your second option) is that we can style/format the result -- I expect to add links and stuff to these at some point to give people more direction. Since we can't really do that if the SDK generates the message, I'd say that one is ruled out.

I think this is still cleaner than my approach, so I'll merge it onto mine.

paulgb (approved):

PR #82, opened 2023-08-22
Abhi/rusty s3 switch

pretentious7:
switches to using rusty s3


paulgb:
Rather than duplicating code on the native side, let's put the common behavior into `core`.

This would include:
- `S3Store` type
- `S3Config` type
- Dependency on `rusty-s3`

pretentious7:
replaced by https://github.com/drifting-in-space/y-sweet/pull/138

PR #83, opened 2023-08-23
Add support for key ID in tokens

paulgb:
This adds support for a concept of "key IDs" in token authentication.

This is a feature of the "core" library only, and is not exposed by this PR in either the native or Cloudflare server.

Key IDs are used by consumers of the core library by calling `with_key_id(key_id)`
on an `Authenticator`, which returns a new `Authenticator` using that key ID.

An `Authenticator` with a key ID has its behavior changed:

- When generating keys, it prepends `{key_id}.`.
- When validating tokens, it expects to find `{key_id}.` as a prefix, which is stripped
  before validating the token as usual.

### DIFF
--- a/crates/y-sweet-core/src/auth.rs
+++ b/crates/y-sweet-core/src/auth.rs
@@ -21,13 +21,15 @@ pub enum AuthError {
     InvalidResource,
     #[error("The token signature is invalid")]
     InvalidSignature,
+    #[error("The key ID did not match")]
+    KeyMismatch,
 }
 
 #[derive(Serialize, Deserialize, PartialEq, PartialOrd, Debug)]
 pub struct Authenticator {
     #[serde(with = "b64")]
     private_key: Vec<u8>,
-    server_token: String,
+    key_id: Option<String>,
 }
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


crates/y-sweet-core/src/auth.rs
```diff
@@ -165,24 +129,81 @@ fn hash(bytes: &[u8]) -> Vec<u8> {
...
+        if expected_token != auth_req.token {
+            Err(AuthError::InvalidSignature)
+        } else if auth_req.payload.expiration_millis.unwrap_or(u64::MAX) < current_time {
+            Err(AuthError::Expired)
+        } else {
+            Ok(auth_req.payload)
+        }
+    }
+
+    pub fn with_key_id(self, key_id: String) -> Self {
```
pretentious7, line 194: does the key id itself not have to be url safe base 64?

crates/y-sweet-core/src/auth.rs
```diff
@@ -165,24 +129,81 @@ fn hash(bytes: &[u8]) -> Vec<u8> {
...
+        } else {
+            if self.key_id.is_some() {
+                return Err(AuthError::KeyMismatch);
+            }
+
+            &token
+        };
+
+        let auth_req: AuthenticatedRequest =
+            bincode_decode(&b64_decode(token)?).map_err(|_| AuthError::InvalidToken)?;
```
pretentious7, line 177: or instead of map err

crates/y-sweet-core/src/auth.rs
pretentious7: ah, it doesn't since it's in the header.

are there any limitations there? or can that be any utf-8

paulgb: It has a restricted alphabet, I guess it doesn’t hurt to check for that. Maybe I’ll restrict to url-safe base64

PR #85, opened 2023-08-23
More fancy cursor demo

paulgb:
Cursors rotate to represent the direction of movement.

https://github.com/drifting-in-space/y-sweet/assets/46173/a382ca36-5443-474f-9080-ea2b1d34f711

### DIFF
--- a/examples/src/app/presence/Presence.tsx
+++ b/examples/src/app/presence/Presence.tsx
@@ -29,7 +29,7 @@ export function randomColor() {
 
 export function Presence() {
   const [myColor, _] = useState(randomColor)
-  const presence = usePresence<Presence>()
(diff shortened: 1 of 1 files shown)


rolyatmax:
Is it a straightforward fix to make cursors disappear when a user disconnects?

paulgb:
> Is it a straightforward fix to make cursors disappear when a user disconnects?

Interestingly, this happens on Firefox but not Chrome! Maybe Chrome allows the client to fire a message when the tab is closed that Firefox does not. In any case, it would also affect a Chrome user whose internet connection suddenly goes out.

I think I can fix it by having the server clear the awareness state when a client connection is dropped, but since it's unrelated to this I'll do it in a follow-up PR.

rolyatmax (approved):
from offline convo

PR #96, opened 2023-08-26
Add version update script

paulgb:
Adds a script for updating npm versions.

This repurposes the script that was used before by the github workflow to check that versions are in sync. (That check still exists, but uses the new script instead.)

If the script is passed no CLI arguments, it will check existing dependencies and return a non-zero value if they are mismatched (same behavior as before).

If the script is passed a CLI argument, that argument should be a valid semver version number. The new behavior is to:

- For each package in the workspace:
    - If the package is not marked as `private`, update its version to the new version number.
    - Update the versions any dependencies that are also in the workspace.
    - Save the new `package.json`
 
To run the script:

```npx ts-node scripts/set-js-version.ts 0.0.6```

Or, to just run a check:

```npx ts-node scripts/set-js-version.ts```


rolyatmax (approved):
Looking good - just two things:

1. The `package.json`s that are generated don't have a trailing newline at the end of the file.
2. The package lock file for the workspace needs to be regenerated after the version bump.

paulgb:
> Looking good - just two things:
> 
> 1. The `package.json`s that are generated don't have a trailing newline at the end of the file.
> 2. The package lock file for the workspace needs to be regenerated after the version bump.

Good point. Re #2, is running `npm install` in root sufficient to do this?

rolyatmax:
> Good point. Re #2, is running `npm install` in root sufficient to do this?

Yup!

PR #98, opened 2023-08-26
Simplify integration test config

paulgb:
We don't need a separate test environment just to test auth, we can just pass in an auth environment variable (thanks @pretentious7 for figuring that out!)


crates/y-sweet-worker/wrangler.toml
```diff
@@ -47,17 +47,3 @@ bindings = [
 binding = "Y_SWEET_DATA"
 bucket_name = "y-sweet-data-staging"
 preview_bucket_name = "y-sweet-data-dev"
-
-[env.auth-test]
-vars.WORKERS_RS_VERSION = "0.0.11"
```
pretentious7, line 52: this variable not needed anymore?

paulgb: the entire `auth-test` environment is no longer needed. Instead we just use the test env and pass an additional variable.

pretentious7: Yea but the version variable is not defined in the env vars

paulgb: It's defined on line 39 for `[env.test]`
image

The code that invoked `[env.auth-test]` no longer exists

image

PR #102, opened 2023-09-01
connection string language update in examples

feliciachang:


js-pkg/sdk/README.md
```diff
@@ -46,9 +43,9 @@ type HomeProps = {
 
 export default async function Home({ searchParams }: HomeProps) {
     // Point to a local or production y-sweet server.
-    const apiEndpoint = "http://localhost:8080"
+    const connectionString = "http://localhost:8080"
```
rolyatmax, line 46: maybe we change this to something like:
```ts
const connectionString = "[MY_CONNECTION_STRING]"
```
And then mention either in text or in a comment where you can get a connection string.

rolyatmax (approved):
One comment there, otherwise LGTM

PR #107, opened 2023-09-04
Add version to objects

paulgb:
This adds a field called `__version` to the `Y.*` objects returned by `useMap`, `useArray`, and `useText`. This field is incremented only when the underlying document data has changed, enabling it to be used as a parameter to a `useMemo` that acts on the map/array/text object in question.

I also implement the change on the voxel editor, where it should lead to a performance improvement.


js-pkg/react/src/main.tsx
```diff
@@ -161,57 +161,70 @@ export function YDocProvider(props: YDocProviderProps) {
...
-function useRedraw() {
-  const [_, setRedraw] = useState(0)
-  return useCallback(() => setRedraw((x) => x + 1), [setRedraw])
+function useVersion(): [number, () => void] {
+  const [version, setRedraw] = useState(0)
+  return [version, useCallback(() => setRedraw((x) => x + 1), [setRedraw])]
 }
 
-export type ObserverKind = 'deep' | 'shallow' | 'none'
+export type ObserverKind = 'deep' | 'shallow' | 'none' | boolean
```
rolyatmax, line 169: Hmm, what's the upside here? Is this solving a specific problem? If it's just another way to specify `deep` and `none`, I'd vote for removing the boolean. I think it adds ambiguity for readers - makes them have to know what `true` and `false` mean as configuration values alongside strings. It's also potentially confusing for writers who see that there are 5 possible values for this property and either have to look up the difference between them all, or, worse, intuit incorrectly what they mean.

paulgb: good point, removed

rolyatmax:
I think there are some gotchas with this approach:

1. Using `__version` in a dependency array is probably often worse than just using the value in the Y* object. The exception might be where lots of values in a deeply-nested object would otherwise be used.

For example,

```tsx
function MyComponent() {
  const obj = useMap('my-map')
  useEffect(() => {
    // do something with obj.foo
  }, [obj.foo]) // better to use obj.foo here instead of obj.__version since __version may change more often than foo
}
```

If there are lots of values in a deeply-nested object that are used in the hook, then it might be easier to use `__version`, but you could end up with some false positives.

The way I think we're expected to deal with the object-reference-not-changing issue is to only use primitive values (no objects or arrays) in dependency lists. I _think_ the react hooks linter `exhaustive-deps` rule might even complain if you stick an object in the dependency list? Not sure.

Anyway, the point here  is that `__version` is probably not something most people should reach for, since, in my experience, dependency lists can usually just include primitive values without things getting out of control.

2. It could be easy for devs to think that `__version` is some canonical version for the object across all clients and use it as such. Even if devs figure out that gotcha, they might think that a Yjs object's `__version` is at least the same across components, but that's not true either.

For example,

```tsx
function Outer() {
  const textOuter = useText('my-text')
  return someConditional ? <Inner textOuter={textOuter} /> : null
}

function Inner(props) {
  const textInner = useText('my-text')
  props.textOuter.__version === textInner.__version // false
}
```

Two components that reference the same Yjs object can have different version numbers, even if both objects are up to date. That's because the `[version, setVersion] = setState(0)` call that happens deep in the hooks is creating a version that's local to the component the `useText` is used in. (Actually, it's local to the component + order the hook was called in. So two `useText('my-text')`s in the same component will technically have different a piece of version state.)

This ends up being a pretty big gotcha if these objects are passed around to different components and compared, used for debugging, or are used for anything besides dependency arrays, really.

Given all this, I wonder if the downsides are greater than the upsides. Maybe a better solution would be not using React state to store the `__version`? Is it possible to have a version stored on the Yjs object itself, so that it's the same across all clients?

PR #108, opened 2023-09-05
Abhi/remove s3 aws hardcoding

pretentious7:


crates/y-sweet-worker/src/config.rs
```diff
@@ -7,9 +7,10 @@ const BUCKET: &str = "Y_SWEET_DATA";
 const BUCKET_KIND: &str = "BUCKET_KIND";
 const AUTH_KEY: &str = "AUTH_KEY";
 const CHECKPOINT_FREQ_SECONDS: &str = "CHECKPOINT_FREQ_SECONDS";
-const S3_ACCESS_KEY_ID: &str = "AWS_ACCESS_KEY_ID";
-const S3_SECRET_ACCESS_KEY: &str = "AWS_SECRET_ACCESS_KEY";
-const S3_REGION: &str = "AWS_REGION";
+const S3_ACCESS_KEY_ID: &str = "S3_ACCESS_KEY_ID";
```
paulgb, line 10: Let's stick to the original naming; these are canonical names used by AWS and are also used in S3-compatible systems.

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

crates/y-sweet-worker/src/s3_store.rs
```diff
@@ -18,17 +19,23 @@ pub struct S3Store {
...
         aws_secret: String,
     ) -> Self {
         let credentials = Credentials::new(aws_access_key_id, aws_secret);
-        let endpoint = format!("https://s3.dualstack.{}.amazonaws.com", region)
-            .parse()
-            .expect("endpoint is a valid Url");
-        let path_style = rusty_s3::UrlStyle::VirtualHost;
+        let endpoint: Url = endpoint.parse().expect("endpoint is a valid url");
+        let path_style =
+			// if endpoint is localhost then bucket url must be of forme http://localhost:<port>/<bucket>
```
paulgb, line 32: tabs

crates/y-sweet-worker/src/s3_store.rs
```diff
@@ -18,17 +19,23 @@ pub struct S3Store {
 
 impl S3Store {
     pub fn new(
+        endpoint: String,
```
paulgb, line 22: might be easier to just take the `S3Config` object directly? We will want to move both into `crates/y-sweet-core` eventually

crates/y-sweet-worker/src/config.rs
```diff
@@ -10,6 +10,7 @@ const CHECKPOINT_FREQ_SECONDS: &str = "CHECKPOINT_FREQ_SECONDS";
 const S3_ACCESS_KEY_ID: &str = "AWS_ACCESS_KEY_ID";
 const S3_SECRET_ACCESS_KEY: &str = "AWS_SECRET_ACCESS_KEY";
 const S3_REGION: &str = "AWS_REGION";
```
paulgb, line 12: AWS uses `AWS_DEFAULT_REGION`.

crates/y-sweet-worker/src/config.rs
```diff
@@ -10,6 +10,7 @@ const CHECKPOINT_FREQ_SECONDS: &str = "CHECKPOINT_FREQ_SECONDS";
...
+const S3_ENDPOINT: &str = "S3_ENDPOINT";
```
paulgb, line 13: AWS uses `AWS_ENDPOINT_URL`

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

crates/y-sweet-worker/src/config.rs
```diff
@@ -10,6 +10,7 @@ const CHECKPOINT_FREQ_SECONDS: &str = "CHECKPOINT_FREQ_SECONDS";
...
 const S3_BUCKET_PREFIX: &str = "S3_BUCKET_PREFIX";
 const S3_BUCKET_NAME: &str = "S3_BUCKET_NAME";
```
paulgb, line 15: Despite the changes to the above env vars, I think we should leave these as `S3_` because they do not have AWS equivalents.

crates/y-sweet-worker/src/s3_store.rs
```diff
@@ -4,8 +4,11 @@ use reqwest::{Client, StatusCode};
 use rusty_s3::{Bucket, Credentials, S3Action};
 use std::{cell::RefCell, time::Duration};
 use time::OffsetDateTime;
+use url::Url;
 use y_sweet_core::store::Store;
```
paulgb, line 9: nit: rm newline and run `cargo fmt`

crates/y-sweet-worker/src/s3_store.rs
```diff
@@ -17,19 +20,18 @@ pub struct S3Store {
...
-            .expect("endpoint is a valid Url");
-        let path_style = rusty_s3::UrlStyle::VirtualHost;
-        let bucket = Bucket::new(endpoint, path_style, bucket_name, region)
+    pub fn new(config: S3Config) -> Self {
+        let credentials = Credentials::new(config.key, config.secret);
+        let endpoint: Url = config.endpoint.parse().expect("endpoint is a valid url");
+        let path_style =
+            // if endpoint is localhost then bucket url must be of forme http://localhost:<port>/<bucket>
+            // instead of <method>:://<bucket>.<endpoint>
+            if endpoint.host_str().expect("endpoint Url should have host") == "localhost" {
```
paulgb, line 29: Not a big deal since we will only use this for testing, but fwiw this will not work for 127.0.0.1. I sometimes use 127.0.0.1 over `localhost` because it unambiguously refers to IPv4 localhost and not IPv6 localhost.

paulgb (commented):
A few nits, and then good to go.

crates/y-sweet-worker/src/config.rs
pretentious7: It uses both, which one do you think is preferable?

crates/y-sweet-worker/src/config.rs
pretentious7: I was very confused parsing that line in the docs lol, yes that makes sense.

crates/y-sweet-worker/src/config.rs
paulgb: Good point, let's use `AWS_REGION`.

paulgb (approved):

PR #128, opened 2023-09-14
Split out vanilla JS client into own package

paulgb:
Some things I've learned:

- `tsup` doesn't support typescript project references [(ref)](https://github.com/egoist/tsup/issues/647), which is the preferred way of telling typescript to build dependencies first.
- I tried removing `tsup` and replacing with `tsc -b`, which tells typescript to follow build references
- That fixes the build of the libraries themselves, but then we are producing ES modules in `dist/`, which not everything supports (`tsup` built both, but there is no way to get the same behavior with `tsc` directly).
- Jest doesn't support ES modules by default, and you have to turn on an experimental flag in Node to use it, because the underlying Node feature [is experimental](https://jestjs.io/docs/ecmascript-modules).
  - The solution I picked is to switch to `vitest`, which also means we don't need to configure typescript compilation

**This means we will no longer ship CommonJS + ESM, and instead only ship ESM.** I think this is an acceptable thing to do in 2023 but @rolyatmax what do you think?


rolyatmax (approved):

rolyatmax:
I have a strong preference to keep commonjs support. So much of the pain people experience with JS comes from dependencies only supporting one or the other (including some of the pain in this PR), and don't want y-sweet packages looking like the culprit when someone can't get their build settings just right so that things "just work".

Maybe I don't understand clearly enough what we're gaining with this PR though? Why do we need to get typescript project references working? What problem were we running into?

paulgb:
The dependency graph is (examples/demos) -> @y-sweet/react -> @y-sweet/client. With tsup, this is fragile, because it doesn't know that @y-sweet/client needs to be built before @y-sweet/react. So it would try to build them in an arbitrary order, which would lead to errors if @y-sweet/react was built before @y-sweet-client.

This is what caused most of the checks to fail in [16a4416](https://github.com/drifting-in-space/y-sweet/pull/128/commits/16a441646178dac5a441e26f1ae533df5819954a)

image

Typescript itself has capabilities for avoiding this, through "references", but for whatever reason tsup doesn't support that yet.

This also fixes another issue I've had, which is that when I wipe `node_modules` and `dist` dirs, VS Code doesn't recognize imports from workspace packages even though the symlinks are in node_modules and the build itself works.

image

Unlike tsup though, tsc is made for producing artifacts for mjs OR cjs; it does not have first-class support for producing both. [Here is a branch](https://github.com/drifting-in-space/y-sweet/tree/paulgb/try-building-cjs-too) with an approach for also building cjs with tsc (I didn't get that branch working, but I think that was for other reasons; I think I could combine that approach to still create a cjs version).

I'm not opposed to supporting cjs as well, I just wonder if it's worth the extra [fragility/complexity](https://nodejs.org/dist/latest-v20.x/docs/api/packages.html#dual-package-hazard). Node-fetch has been ESM-only for ~3 years and still does 20mm+ downloads/week; here's [Rich Harris](https://twitter.com/Rich_Harris/status/1423305079541620739) on esm-only packages two years ago, and a [long discussion of trade-offs](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

paulgb:
This is only two days old but possibly relevant https://isaacs.github.io/tshy/

Unlike tsup, it's just a [thin wrapper](https://twitter.com/izs/status/1702339922567553030) around tsc.

rolyatmax:
I pulled down the original commit that was causing problems, and it looks like maybe this fixes it? https://github.com/drifting-in-space/y-sweet/pull/132

paulgb:
> I pulled down the original commit that was causing problems, and it looks like maybe this fixes it? #132

Yep, that seems to do the trick. I'll unwind the other changes (since there are still things in some of the commits I'd like to keep)

js-pkg/client/src/websocket.ts
rolyatmax, line 1: Just curious: why did this have to be forked?

PR #138, opened 2023-09-19
factors out s3 store into core

pretentious7:


crates/y-sweet-core/src/store/s3.rs
```diff
@@ -1,17 +1,27 @@
...
+    pub bucket: String,
+    pub region: String,
+    pub bucket_prefix: Option<String>,
+}
 
 const PRESIGNED_URL_DURATION_SECONDS: u64 = 60 * 60;
 pub struct S3Store {
     bucket: Bucket,
-    _bucket_inited: RefCell<bool>,
+    _bucket_inited: RwLock<bool>,
```
paulgb, line 24: can this just be an `AtomicBool`?

paulgb (commented):
I like it! Not done yet but a few comments for now.

I'm still not sure I understand the value of checking if the bucket exists on first use -- is it because otherwise we can't tell whether a 404 means that no bucket exists or no key exists?

pretentious7:
https://github.com/drifting-in-space/y-sweet/pull/138

> I like it! Not done yet but a few comments for now.
> 
> I'm still not sure I understand the value of checking if the bucket exists on first use -- is it because otherwise we can't tell whether a 404 means that no bucket exists or no key exists?

For get(), yes, that's correct. But mostly I wanted the Store trait functions to be able to assume the bucket exists, or error out.  The first use thing is just so that S3Store::new is not async.

crates/y-sweet-core/src/store/s3.rs
pretentious7: yep, I've use SeqCst ordering cause errors should be more easily reproducible.

paulgb (approved):

Issue #141, opened 2023-09-22
Client generated doc ID


SebastianStehle:
Hi,

is it possible to create the doc ID from the client? I figure it out how to create a document from my go code. Ideally I would call it in the client directly.

Just a simple API call that creates the doc if not created yet, would be ideal.

At the moment I also get an invalid JSON response:
![image](https://github.com/drifting-in-space/y-sweet/assets/1236435/9360a3f7-bf8d-4f9b-a324-e7437f6dd13b)

paulgb:
Good catch. Yes, we have this now! Will close this, but let me know if anything isn't clear.

PR #142, opened 2023-09-25
Allow specifying document ID

paulgb:
In some cases, a developer may already be managing unique IDs (e.g. sequential or UUIDs) for a set of objects, and would like to map these to y-sweet documents. They could maintain a mapping from their own IDs to y-sweet document IDs, but it's easier for us to just allow them to provide an ID when they create the document.

I think this was suggested by @colelawrence, and it also relates to (but does not close) #141.


.github/workflows/integration-tests.yml
```diff
@@ -45,3 +43,10 @@ jobs:
         Y_SWEET_S3_BUCKET_PREFIX: testing
         Y_SWEET_S3_BUCKET_NAME: ysweet-testing-y-sweet-data
       working-directory: tests
+      timeout-minutes: 10
+
+    - uses: actions/upload-artifact@v3
```
paulgb, line 48: this PR adds log redirection to files for debugging (only for the native server, currently). This uploads those to GitHub to make debugging easier.

js-pkg/sdk/src/main.ts
```diff
@@ -214,8 +219,8 @@ export class DocumentManager {
    *
    * @returns A {@link DocCreationResult} object containing the ID of the created document.
    */
-  public async createDoc(): Promise<DocCreationResult> {
-    const result = await this.doFetch('doc/new', { method: 'POST' })
```
paulgb, line 218: This only worked by mistake: `doFetch` sends a POST request if the body is defined, and this sets the body to `{method: POST}`. The server ignores the `method` field.

tests/.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+out
```
paulgb, line 2: Tests now create some output here for debugging.

js-pkg/sdk/src/main.ts
rolyatmax: I think I'd vote for just having `doFetch` accept a method (like `fetch()` does) rather than having to pass an empty object as the body in order to "hack around" the faulty heuristic of setting the method based on the body. Thoughts?

rolyatmax: Also, I think the `getClientToken()` method below is using `doFetch()` incorrectly (passing what looks like FetchOptions as the request body).

paulgb: My general principle for the API is that _every_ POST request should have a valid JSON body, even if it is an empty object. Since `doFetch` is not a general-purpose fetch function but one designed for this API, the GET/POST distinction restricts the degrees of freedom that the caller has to break things.

tests/.gitignore
rolyatmax: nit: can we call this `test-output` or something like that?

tests/src/index.test.ts
```diff
@@ -115,6 +115,22 @@ describe.each(CONFIGURATIONS)(
...
+    })
+
+    test('Reject invalid doc name', async () => {
+      await expect(
+        DOCUMENT_MANANGER.createDoc({
+          doc: 'mydoc123!',
+        }),
+      ).rejects.toThrow('400')
+    })
+
```
rolyatmax, line 133: What should happen when creating a doc with a name that's already been used? Is that a 409 Conflict?

paulgb: I think it should just return the same document. This way document creation is idempotent.

crates/y-sweet-worker/src/lib.rs
```diff
@@ -84,12 +84,22 @@ async fn new_doc_handler(req: Request, ctx: RouteContext<ServerContext>) -> Resu
...
+    mut req: Request,
     mut ctx: RouteContext<ServerContext>,
 ) -> std::result::Result<NewDocResponse, Error> {
     check_server_token(&req, ctx.data.auth()?)?;
 
-    let doc_id = nanoid::nanoid!();
+    let body = req
+        .json::<DocCreationRequest>()
+        .await
+        .map_err(|_| Error::BadRequest)?;
```
rolyatmax, line 95: Will a request fail if an empty object isn't sent in the body?

js-pkg/sdk/src/main.ts
paulgb: > Also, I think the `getClientToken()` method below is using `doFetch()` incorrectly (passing what looks like FetchOptions as the request body).

Good point, here's a fix: https://github.com/drifting-in-space/y-sweet/pull/145

crates/y-sweet-worker/src/lib.rs
paulgb: Yes.

js-pkg/sdk/src/main.ts
rolyatmax: I'm not sure it restricts degrees of freedom for the caller. After all, the caller of `doFetch()` can still break things by forgetting to pass an empty object.

My main issue is that it's just easy for the caller of `doFetch` to make bad assumptions about what `doFetch` will do. At this point in the logic, all the dev cares about are the 3 main components of a REST endpoint: method, path, and request body. Seems to me that designing a `doFetch` interface that looks like that is more likely to result in a dev correctly predicting what the code will do.

paulgb: Are you thinking that the signature should be `doFetch(url, method, body?)`? I guess I could use typescript overloads to ensure that a body is passed if and only if the method is `POST`.

Alternatively, I could use `doGet(url)` and `doPost(url, body)`, although they'd still use a shared underlying `doFetch`, so we could still shoot ourselves in the foot by calling the underlying `doFetch` instead of `doGet` or `doPost`.

rolyatmax: I like the update you pushed

rolyatmax (approved):

PR #144, opened 2023-09-25
Satisfy clippy

paulgb:


.vscode/settings.json
rolyatmax, line 1: This file should be ignored

rolyatmax (approved):

PR #149, opened 2023-09-26
Export ClientToken from @y-sweet/react and @y-sweet/client

### LINKED ISSUE #129: Consider `@y-sweet/react` re-exporting `ClientToken` type
I think it would make sense to export this type, so I can do a `const token = useMemo((): ClientToken => ({ ... }), [...])` in my provider.

I can work around this by also adding the sdk package & importing the type from there.

paulgb:
Closes #129.

- Moves `encodeClientToken`, which is only used on the client, into `@y-sweet/client`.
- Exports `ClientToken` from `@y-sweet/react` and `@y-sweet/client`, so that clients do not have to depend on `@y-sweet/sdk` just for type safety.


rolyatmax (approved):
LGTM - is the plan to follow this with a version bump and publish?

rolyatmax:
Also looks like the debugger expects `decodeClientToken` from the sdk?

PR #152, opened 2023-10-04
Replace base64 library

paulgb:
This replaces our use of `base64` with `data_encoding`, which has a better API.

### DIFF
--- a/crates/y-sweet-core/Cargo.toml
+++ b/crates/y-sweet-core/Cargo.toml
@@ -10,21 +10,21 @@ repository = "https://github.com/drifting-in-space/y-sweet"
 [dependencies]
 anyhow = "1.0.72"
 async-trait = "0.1.71"
-base64 = "0.21.2"
 bincode = "1.3.3"
+data-encoding = "2.4.0"
 getrandom = { version = "0.2.10", features = ["js"] }
 lib0 = "0.16.9"
 rand = "0.8.5"
+reqwest = "0.11.18"
 rusty-s3 = "0.4.1"
 serde = { version = "1.0.173", features = ["derive"] }
 serde_json = "1.0.103"
 sha2 = "0.10.7"
 thiserror = "1.0.44"
+time = { version = "0.3.25", features = ["wasm-bindgen"] }
 tracing = "0.1.37"
 yrs = "0.16.9"
 yrs-kvstore = "0.1.0"
-reqwest = "0.11.18"
-time = { version = "0.3.25", features = ["wasm-bindgen"] }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


crates/y-sweet-core/Cargo.toml
```diff
@@ -10,21 +10,21 @@ repository = "https://github.com/drifting-in-space/y-sweet"
 [dependencies]
 anyhow = "1.0.72"
 async-trait = "0.1.71"
-base64 = "0.21.2"
 bincode = "1.3.3"
+data-encoding = "2.4.0"
 getrandom = { version = "0.2.10", features = ["js"] }
 lib0 = "0.16.9"
 rand = "0.8.5"
+reqwest = "0.11.18"
```
paulgb, line 18: `reqwest` and `time` are merely moved into the sorted order here, the imports are the same.

crates/y-sweet-core/src/auth.rs
```diff
@@ -1,18 +1,56 @@
...
+/// let spec = Specification {
+///     ignore: "=".to_string(),
+///     translate: Translate {
+///         from: "/+".to_string(),
+///         to: "_-".to_string(),
+///     },
+///     ..BASE64URL_NOPAD.specification()
+/// };
+/// use y_sweet_core::auth::BASE64_CUSTOM;
+/// assert_eq!(BASE64_CUSTOM, spec.encoding().unwrap());
```
paulgb, line 24: This assert is tested by `cargo test`, which ensures that if the code above is updated, the encoding below needs to be too.

rolyatmax: Oh interesting! Would it be too much of a hassle to include this explanation as a comment in the code?

crates/y-sweet-core/src/auth.rs
```diff
@@ -1,18 +1,56 @@
...
+/// ```
+pub const BASE64_CUSTOM: Encoding = Encoding::internal_new(&[
```
rolyatmax, line 26: What's the reason for inlining this, rather than having this in our code?

```rs
let spec = Specification {
    ignore: "=".to_string(),
    translate: Translate {
        from: "/+".to_string(),
        to: "_-".to_string(),
    },
    ..BASE64URL_NOPAD.specification()
};
// ...
spec.encoding().unwrap()
```

paulgb: Just performance. This would run every time we encoded/decoded a string (at least once per request) and involves at least three heap allocations (one for each to_string).

It’s arguably a premature optimization, but it is a low-cost one that I’m happy to make now.

rolyatmax (approved):

PR #156, opened 2023-10-09
don't try to set cache property on fetch on cloudflare worker in sdk

zwily:


rolyatmax:
@zwily Thanks for this PR! I've been spending some time today learning about how the Nextjs and Cloudflare `fetch()` functions cache responses. The more I learn about the complexities of each API, the more I'm thinking that maybe this problem can be solved by including `cache-control: no-store` and `expires: 0` headers in the response from the y-sweet server. If Nextjs and Cloudflare honor these headers when determining whether or not it should cache, then I think I might prefer that approach over adding logic in the SDK for each flavor of `fetch()` we encounter.

From [these docs](https://developers.cloudflare.com/workers/examples/cache-using-fetch#caching-html-resources), it looks like Cloudflare Workers will honor the TTL set in the response headers (assuming they're referring to the `expires` header). As for Nextjs, I'm not entirely sure from just reading [their](https://nextjs.org/docs/app/building-your-application/caching#data-cache) [docs](https://nextjs.org/docs/app/api-reference/functions/fetch).

zwily:
@rolyatmax I agree, that would be better. Good luck in your investigations. :)

rolyatmax:
@zwily I think this approach might be the easiest right now - https://github.com/drifting-in-space/y-sweet/pull/158

rolyatmax:
@zwily #158 has landed and been published in `sdk@0.0.9`. I'm going to close this PR for now. Let me know if that change works for you!

PR #159, opened 2023-10-12
bump js packages to v0.0.9

rolyatmax:
This also updates the `set-js-version` script to only run `npm install` after all the package versions have been updated. Running npm install after each individual update isn't necessary and will also occasionally fail since it updates a package's dependencies to the new version before the dependencies themselves have the new version in their `package.json`s.

This also removes the debugger's package-lock file which is unused when using workspaces. (All packages that are part of the same workspace, share a package-lock file.)


paulgb (approved):

rolyatmax:
Published v0.0.9 of sdk, react, and client

PR #164, opened 2023-10-20
Moving dev docs to github repo [DIS-1267]

feliciachang:


rolyatmax:
Thinking a bit more about this...

When we replace y-sweet.dev with the repo itself, the repo's README effectively becomes the landing page. That page is basically already the same as the current y-sweet.dev introduction page. One thing it's missing though: the Hosted Y-Sweet section. I think we should have that in there.

Once we've done that, we'll also want to add links from the repo README to the other parts of the docs that we're now just keeping in Github. For the "running y-sweet" link, we can link to the `running.md` file (note: we should rename that from mdx -> md, so it renders correctly in github). Then, the links that point to the `react`, `sdk`, and `client` API reference should probably just be the READMEs in those `js-pkg` directories (e.g. `js-pkg/react/README.md`) instead of creating new `docs/react.md` files (since they are basically just copies).

I think that means the necessary changes to this PR are:

1. Rename `docs/running.mdx` to `docs/running.md`.
2. Delete `docs/react.mdx` and `docs/sdk.mdx` files.
3. Update the repo's `README.md` to include the Hosted Y-Sweet section.
4. Update the repo's `README.md` docs section links to point to `docs/running.md`, `js-pkg/react/README.md`, `js-pkg/client/README.md`, and `js-pkg/sdk/README.md`.

What do you think about all that?

feliciachang:
@rolyatmax most of this sounds good to me - i went and updated the old docs in js-pkg with the newer content in react.mdx and sdk.mdx. 

One thing about updating the links in the README - the docs they're linking to right now are the generated docs that paul made. I'd thought that we were planning to keep those?

rolyatmax:
> One thing about updating the links in the README - the docs they're linking to right now are the generated docs that paul made. I'd thought that we were planning to keep those?

@feliciachang Ah yes, that makes sense to me! This PR LGTM!

rolyatmax (approved):

PR #171, opened 2023-10-20
optionally read server settings from env variables

Gentle:
Clap can read Args from env variables, which is better than having them as command line parameters and is optional anyway

### DIFF
--- a/crates/y-sweet/Cargo.toml
+++ b/crates/y-sweet/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/drifting-in-space/y-sweet"
 anyhow = "1.0.72"
 async-trait = "0.1.71"
 axum = { version = "0.6.19", features = ["ws", "headers"] }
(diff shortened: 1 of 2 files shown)


paulgb:
Thanks, will take a look shortly!

Gentle:
I added a variable for store too now, so you can now set everything (except --prod) from environment variables

the naming is open for discussion, Y_SWEET_CHECKPOINT_FREQ_SECONDS is quite a mouthful

paulgb:
> I added a variable for store too now, so you can now set everything (except --prod) from environment variables
> 
> the naming is open for discussion, Y_SWEET_CHECKPOINT_FREQ_SECONDS is quite a mouthful

I think this is fine, it makes sense to stick to the convention. I suspect these will normally not be typed out by hand but used in scripts, so the length is not a big issue.

paulgb (approved):

PR #184, opened 2023-10-31
Check that we can parse response

paulgb:
This should fix the bug @rolyatmax encountered.


rolyatmax:
@paulgb looks like some tests are failing?

rolyatmax (approved):
Pulled down this branch and ran tests locally, which passed. Not sure how to deal with these integration test failures without more error logs (from the github action) to indicate what actually went wrong in the code. 🤔

PR #186, opened 2023-11-06
change copy link to copy link and open it in a new tab

feliciachang:


examples/src/components/CopyLink.tsx
```diff
@@ -1,33 +1,46 @@
...
+      <span className="pr-2">
+        To simulate another user appearing on this document, copy the link to this document and open
+        it in a new window. When you interact on one screen, you should see the user action happens
+        on another.
+      </span>
       <button
-        className="flex text-sm items-center gap-1 px-2 py-1 rounded-lg bg-neutral-50 border-white border transition-all hover:bg-white"
-        onClick={copyLinkToClipboard}
+        className="flex items-center gap-1 px-4 py-2 mt-4 rounded-lg bg-pink-950 text-white border transition-all "
+        onClick={openLinkInNewTab}
```
rolyatmax, line 40: If we change this `button` to an anchor tag, then we can get rid of the `openLinkInNewTab` function altogether. It'll also make for more semantic / accessible markup that works better with screen readers. And we can style it the same as the button!

examples/src/components/CopyLink.tsx
```diff
@@ -1,33 +1,46 @@
...
+      <div className="flex justify-between pb-2">
+        <Title>How to collaborate on this document</Title>
+        <button onClick={() => setHideCallout(true)}>
+          <XMarkIcon className="h-5 w-5 hover:text-black" />
+        </button>
+      </div>
+      <span className="pr-2">
+        To simulate another user appearing on this document, copy the link to this document and open
+        it in a new window. When you interact on one screen, you should see the user action happens
+        on another.
```
rolyatmax, line 36: Another option:

"To simulate collaborating on this document, open this page in a new window. When you make edits in one window, you should see the document updated in the other."

examples/src/app/layout.tsx
```diff
@@ -12,7 +13,8 @@ export default function RootLayout({ children }: { children: React.ReactNode })
     <html lang="en">
       <body className="bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-pink-50 to-pink-900">
         <Sidebar>
-          <div className="absolute w-full h-full top-0 pt-14 lg:relative lg:w-auto lg:pt-0 bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-white/90 via-pink-50/90 to-pink-100/90 sm:mr-2 sm:rounded-lg">
+          <CopyLink />
+          <div className="overflow-auto absolute w-full h-full top-0 pt-14 lg:relative lg:w-auto lg:pt-0 bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-white/90 via-pink-50/90 to-pink-100/90 sm:mr-2 sm:rounded-lg">
```
rolyatmax, line 17: This ends up appearing in the table of contents, too, so maybe we hide it from that page?

Screenshot 2023-11-06 at 4 39 16 PM

rolyatmax (commented):
Some comments there, but otherwise, this is looking good!

examples/src/components/CopyLink.tsx
```diff
@@ -1,34 +1,28 @@
...
-        <LinkIcon className="h-3 w-3" />
-        {copied ? 'Copied!' : 'Copy Link'}
-      </button>
+    <div className="mr-2 mb-2 text-left items-center text-neutral-500 border-2 border-yellow-200 rounded-lg bg-yellow-50 px-6 py-3">
+      <div className="flex justify-between items-center pb-1">
+        <Title>Collaborate on this document</Title>
+        <div className="flex">
+          <a
+            className="text-sm flex items-center gap-1 px-3 py-1 rounded-lg bg-pink-950 text-white border transition-all "
+            href={window.location.href}
```
rolyatmax, line 14: It looks like this is failing to compile causing [the y-sweet demos test to fail](https://vercel.com/drifting-corp/y-sweet-demos/88R41ACSZu6LAHmqBGP54pRjGDo1).

I think we can get around using `window.location.href` by using Nextjs's `usePathname` hook to get the path of the current page.

examples/src/app/(demos)/layout.tsx
```diff
@@ -0,0 +1,12 @@
+import CopyLink from '@/components/CopyLink'
+
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <>
```
rolyatmax, line 5: I noticed that the container for the demos has been pushed down with the new CopyLink component now above it, but it keeps the height of the window, so it adds extra unneeded scroll to every page. I think we can fix this by wrapping the page in a flexbox container. That would just mean replacing this `<>` with `<div className="h-full flex flex-col">`.

(The Tree CRDT demo sometimes still has extra scroll, but that's because it has a fixed 600x600 size for its SVG element. I think we can leave that as is, though.)

examples/src/app/layout.tsx
```diff
@@ -11,11 +11,7 @@ export default function RootLayout({ children }: { children: React.ReactNode })
   return (
     <html lang="en">
       <body className="bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-pink-50 to-pink-900">
-        <Sidebar>
-          <div className="absolute w-full h-full top-0 pt-14 lg:relative lg:w-auto lg:pt-0 bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-white/90 via-pink-50/90 to-pink-100/90 sm:mr-2 sm:rounded-lg">
```
rolyatmax, line 15: Since we moved this wrapper into the `(demos)` directory, we lost the styles from the TLDraw demo which sits outside `(demos)`.

To fix this, can we add `sm:mr-2 sm:rounded-lg overflow-hidden bg-white` to the `tldraw__editor` div in the `TldrawExample` component in `Tldraw.tsx`?

examples/src/app/page.tsx
```diff
@@ -7,7 +7,7 @@ type HomeProps = {
 
 export default async function Home({ searchParams }: HomeProps) {
   return (
-    <div className="space-y-4 p-4 lg:p-8">
+    <div className="space-y-4 p-4 lg:p-8 relative w-auto bg-[radial-gradient(at_bottom_left,_var(--tw-gradient-stops))] from-white/90 via-pink-50/90 to-pink-100/90 rounded-lg">
```
rolyatmax, line 10: I think we need to add `h-full` here to make sure this container fills the height of the page.

examples/src/components/CopyLink.tsx
```diff
@@ -1,17 +1,21 @@
...
   return (
-    <div className="mr-2 mb-2 text-left items-center text-neutral-500 border-2 border-yellow-200 rounded-lg bg-yellow-50 px-6 py-3">
+    <div className="mb-2 text-left items-center text-neutral-500 border-2 border-yellow-200 rounded-lg bg-yellow-50 px-6 py-3">
       <div className="flex justify-between items-center pb-1">
         <Title>Collaborate on this document</Title>
         <div className="flex">
           <a
             className="text-sm flex items-center gap-1 px-3 py-1 rounded-lg bg-pink-950 text-white border transition-all "
-            href={window.location.href}
+            href={`${pathname}?doc=${params.get('doc')}`}
```
rolyatmax, line 18: Strange, but this doesn't appear to be working for me. When I click the button, it opens a new window with the right path, but the query string is `doc=null`. 🤔

rolyatmax: Ahhh - it looks like this bug might be because [shared layouts aren't re-rendered after a navigation](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#3-partial-rendering). So whatever the search params were when the user landed on the site is what they'll be stuck as :-(

rolyatmax (approved):
All this looks good - just one tiny comment left to address: https://github.com/drifting-in-space/y-sweet/pull/186#pullrequestreview-1721213032

I'll go ahead and approve, so you can feel free to land this once that's fixed! Nice work on this!

PR #188, opened 2023-11-21
y-sweet vanillajs example

### LINKED ISSUE #187: Simple example for vanilla JS usage
It'd be really helpful for me if there were a simple, end-to-end "hello world" type example which is basically:

1. Run this command to start your server (just add s3 + credentials, or whatever)
2. Create this `index.html` with this JS code (just imported from jsdelivr/unpkg for this demo - no npm/build/etc.)
3. Create this simple `index.js` and run `node index.js`

And now you simply open `http://localhost:3000?room=75hdo3uud7ji`, and you can e.g. type in a textarea and other people in the same room can see/edit it too.

This would help me see how it works from a very simple example, and then I can launch of from there and start fiddling with it, 
 reading through the docs, and explore the other demos.

I don't use Next.js or React, so that's probably the main reason I'm having trouble understanding how it all works, but I also think it's kinda hard to see what `y-sweet` "*is*" from the current readme/docs - i.e. what sort of API does it expose, how do I actually use it. A really simple vanilla JS "hello world" would help me get to the "aha!" I think.

rolyatmax:
This moves the existing demos into an `examples/nextjs` directory and adds a VanillaJS example to `examples/vanilla`.

Also, we'll have to update our Vercel configuration that deploys these demos to have it look in the `examples/nextjs` directory. The y-sweet-demos deployment for this PR will fail until then.

closes #187


examples/vanilla/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+This demo shows how you can use Y-Sweet with a VanillaJS project.
+
+This demo requires a connection string for a running Y-Sweet server. You can get one at [app.y-sweet.cloud](https://app.y-sweet.cloud), or by [running the y-sweet server](https://github.com/drifting-in-space/y-sweet/blob/main/docs/running.md) yourself.
+
+## How to run this demo
+
+1. `npm install`
+2. `npm run dev`
+3. In another tab, `CONNECTION_STRING=<your-connection-string> npm run server`
```
paulgb, line 11: nit: "another tab" -> "another terminal", since not all terminals have tabs

examples/vanilla/src/client.js
```diff
@@ -0,0 +1,63 @@
...
+import { createColorGridUI } from './ColorGrid'
+
+const COLORS = ['#500724', '#831843', '#9d174d', '#be185d', '#db2777', '#f472b6', '#f9a8d4', null]
+const GRID_SIZE = 10
+const QUERY_PARAM = 'doc'
+
+async function main() {
+  // First, fetch a client token that can access the docId in the URL.
+  // Or, if the URL does not contain a docId, get a client token for a new doc.
+  const url = new URL('http://localhost:8080/client-token')
```
paulgb, line 12: Should this be :3000? As I understand it, `/client-token` refers to a path on the server served by vite (via `vite src --port 3000`) rather than y-sweet server, which runs on 8080 by default.

paulgb (commented):
Consider adding a simple readme to `examples/` that explains the directories.

examples/vanilla/src/client.js
rolyatmax: This is correct because this demo's server runs on 8080. (Vite only runs a dev server for rebuilding / live-reloading on changes.) I forgot that the y-sweet server runs on 8080 by default, though. Maybe I should change this to 9090 so it doesn't conflict with the y-sweet server when folks choose to run one with these demos.

paulgb: Ah yeah, I ran into that when trying to run it locally.

I think running on 9090 makes sense, so it would be:

Static -> :3000
Express.js server -> :9090
y-sweet -> :8080

paulgb: Actually, weirdly it works for me when I run both servers (express.js and y-sweet) on :8080 (!)

but I think the change is good because I don't know why it works for me

paulgb (approved):

PR #190, opened 2023-12-05
adding code walkthrough in y-sweet readme

feliciachang:


README.md
```diff
@@ -42,6 +42,45 @@ Explore our [collaborative examples](https://github.com/drifting-in-space/y-swee
...
 
+## Usage
+
+Check the [vanilla js example](/examples/vanilla/) for more details.
+
+### On the server
+``` js
+import { createDoc } from '@y-sweet/sdk'
+// Get the client token from the y-sweet server. This is basically the user's "password" to edit the "myDoc123" doc.
+app.get('/client-token', async (req, res) => {
```
rolyatmax, line 53: I think it might be simpler / clearer to drop the `app.get()` part and just have something like:

```ts
import { createDoc } from '@y-sweet/sdk'

// Get the client token from the y-sweet server. This is basically the user's "password" to edit the "myDoc123" doc.
const clientToken = await createDoc('myDoc123', CONNECTION_STRING)
```

README.md
```diff
@@ -42,6 +42,45 @@ Explore our [collaborative examples](https://github.com/drifting-in-space/y-swee
...
+  const clientToken = await createDoc('myDoc123')
```
rolyatmax, line 54: This should include a `CONNECTION_STRING` I think?

README.md
```diff
@@ -42,6 +42,45 @@ Explore our [collaborative examples](https://github.com/drifting-in-space/y-swee
...
+})
+```
+
+### On the client
+``` js
+import * as Y from 'yjs';
+import { createYjsProvider } from '@y-sweet/client';
+
+// First, fetch a client token that can access the docId in the URL.
+const clientToken = await fetch("http://localhost:9090/client-token").then(r => r.json())
```
rolyatmax, line 65: I kinda think we should leave this out and just have a comment down at `createYjsProvider()` that says something like "this is the clientToken that is generated above in the server code and must be passed to the client". Thoughts?

README.md
feliciachang: i dont think createDoc requres a connection string, so was thinking it would be simpler to drop that entirely and rely on other tutorials in the docs to cover the concept of the connection string (just cause explaining it here would mean a big chunk of text

README.md
feliciachang: i think we could probably go without explaining this too since it should probably be implied that the client token would come from the server

README.md
rolyatmax: I'm realizing now that `createDoc('id123', CONNECTION_STRING)` only returns a `DocCreationResult`. So we'll need to call `getClientToken(id, CONNECTION_STRING)` right after to get the client token. And because we'd have to pass both a `CONNECTION_STRING`, maybe using the `DocumentManager` API is the clearest / simplest?

e.g.:

```ts
const manager = new DocumentManager(CONNECTION_STRING)
const { doc } = await manager .createDoc('myDoc123')
const clientToken = await manager.getClientToken(doc, {})
```

feliciachang: just added this lmk what you think!

README.md
```diff
@@ -42,6 +42,46 @@ Explore our [collaborative examples](https://github.com/drifting-in-space/y-swee
...
+
+### On the server
+``` js
+import { createDoc } from '@y-sweet/sdk'
+
+// Pass in a CONNECTION_STRING, which you can get from running npx y-sweet@latest serve locally or from y-sweet cloud
+const manager = new DocumentManager(CONNECTION_STRING)
+
+// Create a doc
+const { doc } = await manager.createDoc('myDoc123')
```
rolyatmax, line 57: The argument passed to `createDoc` should be `{ doc: 'myDoc123' }`

README.md
```diff
@@ -42,6 +42,46 @@ Explore our [collaborative examples](https://github.com/drifting-in-space/y-swee
...
+```
+
+### On the client
+``` js
+import * as Y from 'yjs';
+import { createYjsProvider } from '@y-sweet/client';
+
+// Create the yjs doc and link it to the y-sweet server:
+const doc = new Y.Doc();
+createYjsProvider(doc, clientToken, { disableBc: true });
```
rolyatmax, line 70: I think we can drop the `{ disableBc: true }` part. That's just an option to turn off the broadcast functionality, but I don't think it's relevant to a vanilla example.

rolyatmax (approved):

PR #191, opened 2023-12-06
y-sweet api/sdk/client updates

rolyatmax:
This PR does a number of things related to the discussion on this issue.

 - [x] renames doc -> docId in the API
 - [x] makes `createDoc` and `manager.createDoc` work the same and clarifies behavior in docstrings
 - [x] replaces `getOrCreateDoc` with `getOrCreateDocAndToken`
 - [x] removes unimplemented `AuthDocRequest` args
 - [x] makes `connectionString` args required and reorders function args as a result
 - [x] replaces `DocCreationRequest` with simple `docId`s for now
 - [ ] update docs
 - [ ] update examples
 - [ ] update tests


js-pkg/client/package.json
```diff
@@ -6,7 +6,7 @@
     "dev": "tsup --watch",
     "format": "prettier --write \"**/*.{ts,tsx,js,jsx}\""
   },
-  "version": "0.0.9",
```
paulgb, line 9: even though technically it's not breaking semver because we are in a `0.` version, maybe we should make this a `0.1.0` release because it makes breaking changes to the interface?

js-pkg/sdk/src/main.ts
```diff
@@ -234,16 +229,20 @@ export class DocumentManager {
   }
 
   /**
-   * Create a new, empty document on the y-sweet server.
+   * Creates a new document on the y-sweet server given an optional docId. If document with given
```
paulgb, line 232: nit: if _a_ document

paulgb (approved):

js-pkg/client/package.json
rolyatmax: I'd be down to move to a 0.1.0

rolyatmax:
From offline convo - integration tests seem unrelated to these changes. Landing for now

PR #194, opened 2023-12-08
Update yrs version

paulgb:
`cargo publish --dry-run` now succeeds


rolyatmax (approved):
LGTM.

Maybe instead of releasing a `0.1.1` we could just update the git tag for `v0.1.0` and then regenerate the release artifacts.

PR #196, opened 2023-12-08
Make Minio tests work on Mac

paulgb:
This PR does some yak shaving from debugging some S3 integration test stuff:
- Makes Minio tests work under Docker on Mac (by using a host binding instead of the host network)
- Updates the version of wrangler
- Updates some cargo dependencies
- Disables `worker` tests in CI/CD after determining that the flakiness is coming from `workerd` on Linux.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.4"
+version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


rolyatmax (approved):
LGTM! Remind me, what's the difference between the `run-with-minio` tests and the `test` script in `tests/package.json`? When would you run the `run-with-minio` tests?

If they're "more complete" or if they are the recommended way to run tests locally (e.g. when a newcomer makes a change and wants to make sure they didn't break anything), then I think we should either mention that in the README or make it look like the way to run tests (e.g. moving it to the root and calling it `test.sh` or having an NPM script in the root called test that calls it?) Thoughts?

PR #211, opened 2024-02-17
Support prefix path for R2 buckets

jakelazaroff:
Finally played around with Y-Sweet and it's really slick! Local testing was a breeze and it was super easy to deploy on Cloudflare Workers.

One area in which the R2 support doesn't seem to have parity with S3 (afaict) is in using a bucket prefix. This PR adds that — slightly awkwardly, as it reuses the `S3_BUCKET_PREFIX` environment variable even though it's not an S3 bucket. Happy to change the PR if you have a better way to do this!

### DIFF
--- a/crates/y-sweet-worker/src/config.rs
+++ b/crates/y-sweet-worker/src/config.rs
@@ -105,7 +105,7 @@ impl TryFrom<&Env> for Configuration {
             auth_key_id: None,
             bucket: BUCKET.to_string(),
             s3_store_config: s3_config,
-            bucket_prefix: None,
+            bucket_prefix: env.var(S3_BUCKET_PREFIX).map(|s| s.to_string()).ok(),
             url_prefix: None,
             timeout_interval,
         })


paulgb:
Cool, thanks Jake! Will take a look soon.

paulgb:
Ok, this looks good! I'm going to LGTM it as is, but I took some notes from my code dive for my future self:

- The schema of the [Configuration](https://github.com/drifting-in-space/y-sweet/blob/1928b0871533d77eaa96716f36c20b5712c7263c/crates/y-sweet-worker/src/config.rs#L37-L45) object pre-dates support for S3 on the Cloudflare worker implementation of Y-Sweet.
- When we added support for S3, it was by dropping in an optional `S3Config` object.
- Both the `Configuration` and `S3Config` objects support an optional `bucket_prefix`. Currently, Y-Sweet never sets the `bucket_prefix` of `Configuration`, but it can be used by code that calls Y-Sweet as a library (including the Y-Sweet Cloud platform).
- If `S3Config` is `Some(_)`, the `bucket_prefix` of `Configuration` is set

This PR is a correct approach, and I think it's the best we can do without a breaking API change. But it does make me realize that _for library users of y-sweet-worker_, our current API is confusing, because it potentially allows `bucket_prefix` to be set in two places and only one of them is used.

A potential follow-up here would be to have a `BucketConfig` enum with `S3(S3Config)` as one option an `R2 {bucket_prefix: Option<String>}` as another. This change would break the API for library consumers.

jakelazaroff:
(Y-)Sweet, thank you!

PR #214, opened 2024-03-12
Don't expect a response to a HEAD request

paulgb:


rolyatmax (approved):

michaellee8 (approved):
lgtm, tried on my self-hosting setup and can confirm it works now

Issue #228, opened 2024-05-29
Question: Deploying without cloudflare


bgervan:
Hello, 

Is it possible to deploy it with docker without cloudflare? Or is it on the roadmap?

paulgb:
Hi Benjamin,

It's possible now with the Docker image: https://github.com/jamsocket/y-sweet/pkgs/container/y-sweet

But there are currently a couple of issues with using that for more than a dev server:
- You can only run one instance, it's not horizontally scalable.
- It doesn't currently unload documents when they are no longer in use.

I'm planning to integrate with https://plane.dev which would solve both of those. It's on the roadmap, but I don't have an ETA yet.

paulgb:
These issues are now fixed thanks to @rbtying's work! There's also now a simple reverse proxy setup that could be adapted for a single-node production setup (https://github.com/jamsocket/y-sweet/tree/main/deploy). I'm closing this since that work is done, but feel free to follow up with other questions.

PR #233, opened 2024-06-28
RFC: Implement clean shutdown and in-memory document GC

rbtying:
I noticed in #228 that the y-sweet binary is currently missing at least best-effort persistence-before-shutdown, and GCing documents out of memory. I understand that y-sweet is currently mostly deployed as CloudFlare workers, so both of those are a bit moot - CloudFlare will kill the process entirely once there aren't any connections for long enough -- but I figured that it might be nice to have something for outside-CloudFlare as well.

This PR implements:

- slightly more verbose errors (rather than throwing away error details, they're propagated to the response)
- a shutdown process which attempts to persist all of the documents prior to kiling the process, skipping checkpoint_freq throttling (there's a comment which suggests this is intended?), prior to finishing shutdown
- a simple garbage collection task per document which looks at the number of copies of the Awareness struct, which is held one per connection + one in the docs map. This is a little janky, but I didn't want to modify `y-sweet-core` with implementation details of the server. The garbage collection task holds the cancellation token for the persistence worker, so in graceful shutdown scenarios it should always persist the current state.


crates/y-sweet/src/server.rs
```diff
@@ -145,14 +271,14 @@ impl Server {
         Ok(self
             .docs
             .get(doc_id)
-            .expect("Doc should exist, we just created it.")
+            .ok_or_else(|| anyhow!("Failed to get-or-create doc"))?
```
paulgb, line 274: Maybe we can use `.context()` from `anyhow` here, so that any error details are propagated?

crates/y-sweet/src/server.rs
```diff
@@ -161,16 +287,19 @@ impl Server {
...
             Ok(())
         }
     }
 
     pub async fn serve(self, addr: &SocketAddr) -> Result<()> {
         let listener = TcpListener::bind(addr).await?;
+        let token = self.cancellation_token.clone();
         let server_state = Arc::new(self);
 
+        let server_state_ = server_state.clone();
```
paulgb, line 301: nit: omit this and inline the call to `.clone()`, i.e.

```rust
.with_state(server_state.clone());
```

crates/y-sweet/src/server.rs
```diff
@@ -101,38 +128,137 @@ impl Server {
...
+        cancellation_token: CancellationToken,
+    ) {
+        let mut last_had_references = std::time::Instant::now();
+
+        loop {
+            tokio::select! {
+                _ = tokio::time::sleep(checkpoint_freq) => {
+                    if let Some(doc) = docs.get(&doc_id) {
+                        let awareness = Arc::downgrade(&doc.awareness());
+                        if awareness.strong_count() > 1 {
```
paulgb, line 177: nit: use `Arc::strong_count()` and avoid the downgrade.

crates/y-sweet/src/server.rs
```diff
@@ -101,38 +128,137 @@ impl Server {
...
     }
 
+    async fn doc_gc_worker(
+        docs: Arc<DashMap<String, DocWithSyncKv>>,
+        doc_id: String,
+        checkpoint_freq: Duration,
+        persistence_cancellation_token: CancellationToken,
+        cancellation_token: CancellationToken,
+    ) {
+        let mut last_had_references = std::time::Instant::now();
```
paulgb, line 170: As I understand the logic here, we're GCing if the doc has not had references for two consecutive checks. If that's the case, I think it would be clearer to follow the logic here if it were just a `u8` called something like `checkpoints_without_refs` with these rules:

- it starts as `0`
- if a doc has exactly 1 strong reference, increment it by `1`
- if a doc has > 1 strong reference, reset it to `0`
- if it reaches `2`, GC the doc

crates/y-sweet/src/server.rs
```diff
@@ -101,38 +128,137 @@ impl Server {
...
+        doc_id: String,
+        cancellation_token: CancellationToken,
+    ) {
+        let mut last_save = std::time::Instant::now();
+
+        // There are two ways to trigger a save: either the dirty
+        // callback is called, or the cancellation token has been
+        // cancelled.
+        let mut is_done = false;
+        let mut interrupted = false;
```
paulgb, line 215: I find the flag variables make it hard to convince myself that this is right. This is psuedocode, but I think it reflects the desired behavior with a more linear flow:

```rust

loop {
    // wait until dirty
    select! {
        _ = recv.recv() => {
            // drop through to sleep
        }
        _ = cancellation_token.cancelled() {
             // no need to save because not dirty
             break
        }
    }

    // we are dirty; wait until throttled
    select! {
        _ = sleep_until_save() {
             save();
        }
        _ =  cancellation_token.cancelled() {
             save();
             break
        }
    }
}

```

paulgb (commented):
Thanks for this! I like the use of the connection token, that seems like a good fit here.

If you rebase on main, the integration tests should pass -- they were breaking because of an unrelated cloudflare workers version issue.

crates/y-sweet/src/server.rs
rbtying: I think this is a style thing - the gc task ~shouldn't be the task keeping the doc alive, and comparing `Arc::strong_count(&doc.awareness()) > 2` seems extra confusing (it requires you to know that internally `doc.awareness` increments the strong count.

If we're willing to change y-sweet-core, we could expose something more purpose-fit for this?

crates/y-sweet/src/server.rs
rbtying: I'll try to poke at this again. The challenge is that we ~can't write `sleep_until_save` without getting into tricky futures/borrows-across-futures issues -- we don't want to move `recv` out of the loop and into a helper func.

how do you feel about manual `poll` implementations here?

crates/y-sweet/src/server.rs
```diff
@@ -39,12 +44,31 @@ fn current_time_epoch_millis() -> u64 {
     duration_since_epoch.as_millis() as u64
 }
 
+#[derive(Debug)]
+pub struct AppError(StatusCode, anyhow::Error);
+impl IntoResponse for AppError {
+    fn into_response(self) -> Response {
+        (self.0, format!("Something went wrong: {}", self.1)).into_response()
```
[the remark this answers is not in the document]
rbtying: done via middleware - I think it's actually preferable to return the relevant error, but it's ~fine to hide it too. Changing it elsewhere (e.g. in From impl) makes it hard to configure at runtime.

crates/y-sweet/src/server.rs
paulgb: Ah, that makes sense, I hadn't considered that `.awareness()` increases the count. I'm fine with changing `y-sweet-core` for this, although I haven't yet thought through what the real ideal API would be. Would exposing something like `.ref_count()` be a reasonable approach?

paulgb (approved):
Looks good, thanks for this! One question/comment, and then I'll merge.

rbtying:
Unfortunately it looks like yrs is not thread-safe (https://github.com/y-crdt/y-crdt/issues/373), so I might tinker a bit to see if there's a way to do the yrs subscription drops single-threaded before merging

rbtying:
> Unfortunately it looks like yrs is not thread-safe ([y-crdt/y-crdt#373](https://github.com/y-crdt/y-crdt/issues/373)), so I might tinker a bit to see if there's a way to do the yrs subscription drops single-threaded before merging

I think this is unblocked w/ the 0.19 update!

paulgb:
> > Unfortunately it looks like yrs is not thread-safe ([y-crdt/y-crdt#373](https://github.com/y-crdt/y-crdt/issues/373)), so I might tinker a bit to see if there's a way to do the yrs subscription drops single-threaded before merging
> 
> I think this is unblocked w/ the 0.19 update!

Merged! Thanks for all your work on getting this across the finish line!

PR #244, opened 2024-07-09
Replace getOrCreateClientAndToken with getOrCreateDocAndToken in README

causztic:
Tried running with the example code in the README and found this undefined method `getOrCreateClientAndToken`

Update README to use the existing `getOrCreateDocAndToken` instead


paulgb (approved):
Thanks for the fix!

PR #245, opened 2024-07-11
Tag docker image with `latest` in build step

paulgb:


rolyatmax (approved):
I'm surprised this doesn't happen automatically! I thought pushing an image without any tags would just give it the `latest` tag

paulgb:
> I'm surprised this doesn't happen automatically! I thought pushing an image without any tags would just give it the `latest` tag

Yeah, it's weird, it uses the branch name by default instead.

PR #254, opened 2024-08-12
Remove unwraps in durable object

paulgb:
We don't get very good error messages currently when errors happen inside the durable object; this fixes that.


feliciachang (approved):
Broadly wondering if some of these error messages can provide more context on why the error is happening but otherwise lgtm!

PR #261, opened 2024-09-03
Fix unit test GitHub Action cargo caching

michaelsilver:
Swatinem/rust-cache@v2 was failing with:
"could not find `Cargo.toml` in `/home/runner/work/y-sweet/y-sweet` or any parent directory"

Use the workspaces feature to point the cache to the correct directory (crates).


michaelsilver:
Verified that this works!

paulgb (approved):
Good catch, thanks!

PR #264, opened 2024-09-04
Implement a `serve-doc` command

michaelsilver:
This will serve an individual document, instead of running a multi-doc server.

Also took the liberty of cleaning up our use of env vars for AWS -- review commit-by-commit for more details. If you disagree with these changes, I can easily drop commits.

### DIFF
--- a/crates/y-sweet/src/main.rs
+++ b/crates/y-sweet/src/main.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use anyhow::Result;
 use clap::{Parser, Subcommand};
 use serde_json::json;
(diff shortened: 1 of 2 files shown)


crates/y-sweet/src/main.rs
```diff
@@ -236,6 +247,76 @@ async fn main() -> Result<()> {
...
+                        }
+                    }
+                } => {
+                    tracing::info!("Received SIGTERM, shutting down.");
+                }
+            }
+
+            cancellation_token.cancel();
+            server_handle.await?;
+            tracing::info!("Server shut down.");
```
rolyatmax, line 294: Is there a way we can make sure everything has been synced to S3 here before completing the shutdown?

michaelsilver: If I'm reading the code correctly, this is already handled in `doc_persistence_worker()`, is that right?

rolyatmax: Ah yeah, I think you're right

rolyatmax (approved):
I think this looks good, though @paulgb might want to take a quick look since he's more familiar with this codebase.

PR #273, opened 2024-09-16
[serve-doc] Make storage config optional

michaelsilver:
Expect both `STORAGE_BUCKET` and `STORAGE_PREFIX` environment variables to be set, or neither. If neither, documents will be stored in memory only. If only one variable is set, panic.


rolyatmax (approved):
This makes sense to me - might be good for @paulgb to take a quick look too since he's more familiar with Y-Sweet's implementation details

PR #274, opened 2024-09-16
Modernize Python SDK packaging

michaelsilver:
I was looking for a way to replace `python -m unittest discover -s tests` with something like `python setup.py test`. As it turns out, python packaging has changed a ton over the last few years; `python setup.py test` is no longer supported:

```
********************************************************************************
Please remove any references to `setuptools.command.test` in all
supported versions of the affected package.

By 2024-Nov-15, you need to update your project and remove deprecated calls
or your builds will no longer be supported.
********************************************************************************
```

So, what is the modern approach to python packaging? Turns out the answer is `pyproject.toml`, summarized here:
https://venthur.de/2022-12-18-python-packaging.html

Even though this might not be the most beautiful thing, it's apparently the path the python community's taken with packaging.

As a result, we can now run tests with simply `pytest`, and keep test/dev deps separate from core deps. I see this change as a step towards making the Python SDK more maintainable for the long term.


paulgb (approved):
Nice! I followed the install/test instructions and they worked.

Issue #269, opened 2024-09-15
y-sweet server npm wrapper cli re-downloads binary every time


kwshi:
I have a relatively slow internet connection and noticed that even after the initial download/install (`pnpm add -D y-sweet@latest`) it always took 30+ seconds for `pnpm exec y-sweet` to launch. Some light digging leads me to think that the binary is getting re-downloaded every time the CLI is invoked, which doesn't seem like the correct/intended behavior.

Here's my train of thought:
- the `js-pkg/server/package.json` specifies the entry point to `pnpm exec y-sweet` as `js-pkg/server/src/run.js`, whose contents are: https://github.com/jamsocket/y-sweet/blob/35ea7d34693ea9a4fd194247945b330c70ba35be/js-pkg/server/src/run.js#L6-L14
- I added some `console.log`s to try to inspect what was happening behind the scenes:
  ```javascript
  async function runBinary() { 
    try { 
      var binpath = await binaryExists()
      console.log('BINARY EXISTS', binpath)
    } catch (e) { // <-- added `(e)` to capture error 
      console.log('BINARY FAILED', e)
      binpath = await installBinary() 
      console.log('INSTALLED BINARY', binpath)
    }
  
    spawnSync(binpath, process.argv.slice(2), { stdio: 'inherit', stderr: 'inherit' }) 
  }
  ```
- The first invocation of `pnpm exec y-sweet` displays `BINARY FAILED Error: Cannot find module 'y-sweet'`, then (after several minutes on my slow internet) `INSTALLED BINARY /tmp/test/node_modules/.pnpm/y-sweet@0.2.2/node_modules/y-sweet/node_modules/y-sweet/bin/y-sweet-0.2.2`. I tried to invoke the binary manually; weirdly enough I had to manually `chmod +x` it first, but then it worked fine. (This happened during my attempts to reproduce the error from a blank project; I don't recall having to do this originally when the error first came up. I don't know what this is about, but it seems like a separate issue.)
- Re-invoking `pnpm exec y-sweet` displays the exact same error message instead of `BINARY EXISTS` etc. and makes me wait several minutes while it redownloads the binary.

In case this information is relevant:
- pnpm 8.15.5
- node 20.15.1

kwshi:
[...]

PR #277, opened 2024-09-17
Fix binary download issue

paulgb:
It was reported in #269 that y-sweet downloads the binary every time. I was able to reproduce this locally when running the script directly, although this may be because `require.resolve` behaves weirdly (fails silently?) outside of an npm package.

I think the underlying issue is that we had two different code paths that determined where to check for the binary and where to put the binary. This collapses them into one code path, and cleans up/modernizes the code in general.


js-pkg/server/src/get-binary.js
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error(`Error downloading ${url}: server returned ${res.status}`)
+  }
+
+  await pipeline(res.body, zlib.createGunzip(), fs.createWriteStream(filePath, { mode: 0o770 }))
+
+  return filePath
+}
+
+exports.getBinary = async () => {
+  let binpath = path.normalize(path.join(__dirname, '..', 'bin'))
```
rolyatmax, line 41: nit: you could replace the `path.normalize(path.join(` with a `path.resolve(`.

rolyatmax (approved):

js-pkg/server/src/get-binary.js
paulgb: TIL! I'm going to go with this though because I've already tested it.

PR #278, opened 2024-09-17
Split up SDK codebase

paulgb:
This splits up the SDK codebase to make it easier to work with. No behavior is changed.


js-pkg/gen-docs/src/main.ts
```diff
@@ -37,7 +37,7 @@ async function generateDocumentation() {
       path.join(jsPkgDir, 'react', 'src', 'main.tsx'),
     ]),
     generateDocumentationForProject(path.join(jsPkgDir, 'sdk'), [
-      path.join(jsPkgDir, 'sdk', 'src', 'main.ts'),
+      path.join(jsPkgDir, 'sdk', 'src', 'index.ts'),
```
rolyatmax, line 40: Is there a reason we're changing `main` to `index`? And just for this one package?

rolyatmax (approved):
Nothing blocking, though I'd recommend we either keep `main` as the entrypoint for all the js packages in `js-pkg` or we change _all_ of them to `index`.

js-pkg/gen-docs/src/main.ts
paulgb: Discussed offline, changed back.

PR #284, opened 2024-09-19
Add release helper

paulgb:
This PR introduces a Rust program that helps with keeping versions in sync across our rust/python/JS codebases, both in the repo and in their respective public registries.

See `releaser/README.md` for instructions on usage. This replaces the [notion checklist](https://www.notion.so/jamsocket/Y-sweet-release-checklist-e2d9fc60c0c143a2aa8ad2ddd4964e73?pvs=4) (note: the current checklist is only accessible to Jamsocket notion workspace members) we previously followed, as well as `set-js-version.ts` which was removed in #285.

Instructions are in the [README](https://github.com/jamsocket/y-sweet/blob/paulgb/releaser/releaser/README.md).

The goal is to reduce the chances of human error and reduce the toil that comes from managing multiple packages across several registries.


releaser/src/package_manager/python/virtualenv.rs
```diff
@@ -0,0 +1,75 @@
...
+        &self.python
+    }
+}
+
+impl Drop for VirtualEnv {
+    fn drop(&mut self) {
+        println!("Removing virtual environment at: {}", self.path.display());
+        std::fs::remove_dir_all(self.path.clone()).ok();
+    }
+}
```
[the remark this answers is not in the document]
paulgb: Can't propagate an error from a drop handler.

releaser/src/git.rs
```diff
@@ -0,0 +1,69 @@
...
+        Ok(())
+    }
+
+    /// Push the current branch to the remote repository.
+    pub fn push(&self) -> Result<()> {
+        Command::new("git")
+            .args(["push", "--set-upstream", "origin", "main"])
+            .status()?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
paulgb: N/A, this code only needs to work for one repo that uses `main`.

releaser/src/package_manager/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl PackageType {
+    pub fn get_package_manager(&self) -> Box<dyn PackageManager> {
+        match self {
+            PackageType::Cargo => Box::new(cargo::CargoPackageManager),
+            PackageType::Node => Box::new(node::NodePackageManager),
+            PackageType::Python => Box::new(python::PythonPackageManager),
+        }
+    }
+}
```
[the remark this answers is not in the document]
paulgb: they do not have a constructor

releaser/src/package_manager/cargo.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn publish(&self, path: &Path) -> Result<()> {
+        let status = Command::new("cargo")
+            .arg("publish")
+            .current_dir(path)
+            .status()?;
+        if !status.success() {
+            return Err(anyhow::anyhow!("Failed to publish package"));
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
paulgb: the cargo binary does this automatically now.

releaser/src/git.rs
```diff
@@ -0,0 +1,69 @@
...
+        let branch_ref = self.repo.branch(branch, &commit, false)?;
+        self.repo.set_head(branch_ref.get().name().unwrap())?;
+        self.repo.checkout_tree(commit.as_object(), None)?;
+        Ok(())
+    }
+
+    /// Push the current branch to the remote repository.
+    pub fn push(&self) -> Result<()> {
+        Command::new("git")
+            .args(["push", "--set-upstream", "origin", "main"])
```
rolyatmax, line 65: Don't we want this to push to the release branch not `main`? I'd expect pushes to `main` to fail

releaser/src/releaser.rs
```diff
@@ -0,0 +1,227 @@
...
+            };
+            println!(
+                "Updating dependencies for {} package {}",
+                style(&package.package_type).bold().red(),
+                style(&package.name).bold().cyan()
+            );
+            let result = package
+                .package_type
+                .get_package_manager()
+                .update_dependencies(&package.path, deps, &bump_version)
```
rolyatmax, line 97: We also need to update `debugger/package.json`, `examples/nextjs/package.json`, and `examples/vanilla/package.json`

releaser/src/git.rs
```diff
@@ -0,0 +1,70 @@
...
+        self.repo.checkout_tree(commit.as_object(), None)?;
+        Ok(())
+    }
+
+    /// Push the current branch to the remote repository.
+    pub fn push(&self) -> Result<()> {
+        let branch = self.get_branch()?;
+        Command::new("git")
+            .args(["push", "--set-upstream", "origin", &branch])
+            .status()?;
```
rolyatmax, line 67: When we push, we should push `--tags`, too

releaser/src/releaser.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        // Check out a branch
+        let branch_name = format!("release/{}", bump_version);
+        self.git.checkout_new_branch(&branch_name)?;
+
+        // Commit changes
+        self.git
+            .commit_all(&format!("Bump version to {}", bump_version))?;
+
```
rolyatmax, line 140: We should also add a git tag to this commit like `v0.3.0`

rolyatmax:
Left some comments re: some steps that were missed from this doc: https://www.notion.so/jamsocket/Y-sweet-release-checklist-e2d9fc60c0c143a2aa8ad2ddd4964e73

releaser/src/releaser.rs
paulgb: The tag is automatically applied when we release, as I understand (e.g. I used this script to deploy v0.3.0 and it shows up here https://github.com/jamsocket/y-sweet/tags)

rolyatmax (approved):

PR #285, opened 2024-09-19
Bump version to 0.3.0

paulgb:
Note: this was created with a WIP tool that keeps versions in check across our rust/python/npm packages, which can be found in the [paulgb/releaser](https://github.com/jamsocket/y-sweet/pull/284) branch.


.github/workflows/check-ts.yml
```diff
@@ -49,10 +49,3 @@ jobs:
     - name: Check Formatting of doc generator
       run: npx prettier --check "**/*.{ts,tsx}"
       working-directory: js-pkg/gen-docs/
-
-    - name: Check Formatting of scripts
-      run: npx prettier --check "**/*.{ts,tsx}"
-      working-directory: scripts/
-
-    - name: Check that workspace package versions match
-      run: npx tsx scripts/set-js-versions.ts
```
rolyatmax, line 58: This seemed like a good check to have. If `set-js-versions` is getting replaced with something else, can we work that back into this repo - maybe as an npm script?

paulgb: my approach has been to move the validation of the state into the release process script, rather than having them checked at every commit. Otherwise, we end up having two sources of truth for what the package layout is etc.

rolyatmax: Gotcha - that makes sense. Since we're losing `set-js-versions`, can we at least add some instructions on where to find and how to run the release process script?

paulgb: When https://github.com/jamsocket/y-sweet/pull/284 is ready, my plan is to replace [these docs](https://www.notion.so/jamsocket/Y-sweet-release-checklist-e2d9fc60c0c143a2aa8ad2ddd4964e73?pvs=4) with a pointer to its readme, which will cover the entire release process (to the extent that it isn't baked into the tool).

rolyatmax (approved):

PR #292, opened 2024-09-26
y-sweet js client support auth endpoint

rolyatmax:
This takes a very crude approach to get the YSweetProvider supporting reconnects. It essentially wraps the `new YSweetProvider` instantiation logic with an async function that calls the user-provided auth endpoint and then instantiates the YSweetProvider. It attaches an `onFailure` handler to the provider allowing it to signal that it needs to be recreated.

A tried several approaches, but found this to be the least invasive. It also - believe it or not - is the most straightforward to understand. It does, however, come with the downsides that the `createYjsProvider` changes in two ways:

1. it's now async
2. it returns an object that is essentially a shimmed Provider that implements most of the Provider's interface
3. the Provider itself now no longer implements the Observable interface, instead it has an `observable` property which gets transferred to the next Provider in case the old one gets recreated

I don't _think_ these are huge downsides as there don't seem to be a lot of use cases where the application needs to grab the provider and do something with it. In the case that an application _does_ need to user the returned provider, then the calling code will need to `await`.

A more robust solution might take the approach of keeping the same Provider but resetting the relevant bits of internal state when getting a new client token from the auth endpoint. That approach ended up being quite messy and error prone with old websocket event handlers and setTimeouts hanging around. It might make sense to clean that up when we rewrite the YSweetProvider in the coming weeks.


paulgb:
> A more robust solution might take the approach of keeping the same Provider but resetting the relevant bits of internal state when getting a new client token from the auth endpoint. That approach ended up being quite messy and error prone with old websocket event handlers and setTimeouts hanging around. It might make sense to clean that up when we rewrite the YSweetProvider in the coming weeks.

I agree with this general approach, the less deeply integrated this is in to the existing provider, the better.

js-pkg/react/src/main.tsx
```diff
@@ -177,33 +186,44 @@ export type YDocProviderProps = {
...
-      disableBc: true,
-    })
 
-    setCtx({ doc, provider, clientToken })
+    ;(async () => {
+      provider = await createYjsProvider(doc, docId, authEndpoint, {
+        // TODO: this disables local cross-tab communication, which makes
+        // debugging easier, but should be re-enabled eventually
+        disableBc: true,
+      })
```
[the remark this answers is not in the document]
rolyatmax: @paulgb what do you think?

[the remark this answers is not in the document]
paulgb: I'm fine with hard-coding it to off for now, we can revisit when we work on the protocol.

examples/nextjs/Dockerfile
```diff
@@ -47,4 +47,4 @@ ENV PORT 3000
 
 # server.js is created by next build from the standalone output
 # https://nextjs.org/docs/pages/api-reference/next-config-js/output
-CMD HOSTNAME="0.0.0.0" node server.js
```
paulgb, line 50: nit: intentional?

js-pkg/client/src/provider.ts
```diff
@@ -469,4 +490,138 @@ export class YSweetProvider extends Observable<string> {
...
+    },
+    get wsconnecting() {
+      return _provider.wsconnecting
+    },
+    get bcconnected() {
+      return _provider.bcconnected
+    },
+    get disableBc() {
+      return _provider.disableBc
+    },
```
paulgb, line 610: For methods like this that change the state of the provider, I think we should just omit them, otherwise they won't behave (the state will reset when we reconnect).

If they are necessary so that we implement an interface, maybe we should throw an error with an “unimplemented” message.

js-pkg/client/src/provider.ts
```diff
@@ -469,4 +490,138 @@ export class YSweetProvider extends Observable<string> {
...
+      return _provider.url
+    },
+    get roomname() {
+      return _provider.roomname
+    },
+    get doc() {
+      return _provider.doc
+    },
+    get awareness() {
+      return _provider.awareness
```
paulgb, line 597: things might get weird if the developer holds the awareness across a new provider reconnect. We might need to mock out that class as well, as ugly as that is :/

js-pkg/client/src/provider.ts
rolyatmax: I agree re: methods that change the state of the provider. (I think `set synced()` is the only one, though? I'll drop that one.)

js-pkg/client/src/provider.ts
rolyatmax: Since we can pass in `awareness` when creating the provider, I'll just use that functionality to keep it referentially stable. See the last commit I pushed there.

paulgb (approved):

PR #299, opened 2024-10-02
Add validity time option to the API (DIS-2749)

### LINKED ISSUE #309: Allow the token validity time to be specified
Currently it is hard-coded, but it should be overridable via the API.

paulgb:
Adds `valid_for_seconds` to the auth body request. We weren't actually parsing the request body before, so we do that now.

Introduces a newtype to avoid mixing up the current time and expiration times.

Closes #309

### DIFF
--- a/crates/y-sweet-core/src/api_types.rs
+++ b/crates/y-sweet-core/src/api_types.rs
@@ -34,7 +34,7 @@ pub struct AuthDocRequest {
     pub user_id: Option<String>,
     #[serde(default)]
     pub metadata: HashMap<String, Value>,
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "validForSeconds")]
     pub valid_for_seconds: Option<u64>,
 }
 
--- a/crates/y-sweet-core/src/api_types.rs
+++ b/crates/y-sweet-core/src/api_types.rs
@@ -34,6 +34,19 @@ pub struct AuthDocRequest {
     pub user_id: Option<String>,
     #[serde(default)]
     pub metadata: HashMap<String, Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub valid_for_seconds: Option<u64>,
+}
+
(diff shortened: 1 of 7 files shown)


tests/src/index.test.ts
```diff
@@ -280,6 +280,44 @@
...
+      test.only('Connecting with 5 validForSeconds should work briefly', async () => {
+        const docResult = await DOCUMENT_MANANGER.createDoc()
+        const conn = await DOCUMENT_MANANGER.getDocConnection(docResult, { validForSeconds: 5 })
+
+        const update = await conn.getAsUpdate()
+        expect(update).toBeDefined()
+
+        // wait 8 seconds
+        await new Promise((resolve) => setTimeout(resolve, 8_000))
+
```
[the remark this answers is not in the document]
paulgb: The process we are testing runs outside of the node runtime, and relies on the system time.

crates/y-sweet-core/src/api_types.rs
```diff
@@ -34,6 +34,19 @@ pub struct AuthDocRequest {
     pub user_id: Option<String>,
     #[serde(default)]
     pub metadata: HashMap<String, Value>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "validForSeconds")]
+    pub valid_for_seconds: Option<u64>,
```
rolyatmax, line 38: Even though this value doesn't exactly make sense in a Jamsocket context, I suppose we should at least allow it to be a part of the request body (and probably log if we see it, too)

rolyatmax (approved):

crates/y-sweet-core/src/api_types.rs
paulgb: If we expose the same parameter in Plane, we could make it work there, too. Plane has an equivalent internally.

rolyatmax: https://linear.app/drifting/issue/DIS-2795/[platform]-allow-y-sweet-auth-doc-endpoint-to-take-valid-for-seconds

PR #330, opened 2024-11-07
Allow doc creation with initial content

michaelsilver:
Closes https://github.com/jamsocket/y-sweet/issues/310

### DIFF
--- a/js-pkg/sdk/src/main.ts
+++ b/js-pkg/sdk/src/main.ts
@@ -135,6 +135,18 @@ export class DocumentManager {
     const clientToken = await this.getClientToken(docId, authDocRequest)
     return new DocConnection(clientToken)
   }
+
+  /**
+   * Creates a new document with initial content.
+   *
+   * @param update A Yjs update byte string representing the initial content.
+   * @returns A {@link DocCreationResult} object containing the ID of the created document.
+   */
+  public async createDocWithContent(update: Uint8Array): Promise<DocCreationResult> {
+    const result = await this.createDoc()
+    await this.updateDoc(result.docId, update)
+    return result
+  }
 }
 
 /**
--- a/tests/src/index.test.ts
+++ b/tests/src/index.test.ts
@@ -341,5 +341,27 @@ describe.each(CONFIGURATIONS)(
         await expect(result).rejects.toContain('401')
       })
     }
+
+    test('Create doc with initial content', async () => {
+      const doc = new Y.Doc()
+      const map = doc.getMap('test')
+      map.set('initial', 'content')
+
+      const update = Y.encodeStateAsUpdate(doc)
+      const docResult = await DOCUMENT_MANANGER.createDocWithContent(update)
+
+      // Verify the doc was created
+      expect(typeof docResult.docId).toBe('string')
+
+      // Connect to the doc and verify the content
+      const getClientToken = async () => await DOCUMENT_MANANGER.getClientToken(docResult)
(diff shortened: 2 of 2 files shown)


js-pkg/sdk/src/main.ts
```diff
@@ -187,3 +199,19 @@ export async function createDoc(
...
+ * @param update A Yjs update byte string representing the initial content.
+ * @returns A {@link DocCreationResult} object containing the ID of the created document.
+ */
+export async function createDocWithContent(
+  connectionString: string,
+  update: Uint8Array,
+): Promise<DocCreationResult> {
+  const manager = new DocumentManager(connectionString)
+  return await manager.createDocWithContent(update)
+}
```
paulgb, line 217: I'm actually tempted to deprecate these functions that sit outside of the `DocumentManager`, so I'm ok with just making this a function on the document manager.

paulgb (commented):
Looks good, can you add a test as well? I know it's a bit pedantic for a three-line function, but:

1. When we make this a single server-side call, it will be nice to have a test that ensures it is compatible with the old approach.
2. It would be a good way to get you familiar with our test setup.

js-pkg/sdk/src/main.ts
```diff
@@ -135,6 +135,18 @@ export class DocumentManager {
...
     return new DocConnection(clientToken)
   }
+
+  /**
+   * Creates a new document with initial content.
+   *
+   * @param update A Yjs update byte string representing the initial content.
+   * @returns A {@link DocCreationResult} object containing the ID of the created document.
+   */
+  public async createDocWithContent(update: Uint8Array): Promise<DocCreationResult> {
```
rolyatmax, line 145: We allow users to create documents with a given ID, but we aren't taking a `docId` param here. I expect we'll have users who want to do this since [the cleanest way to implement the auth endpoint](getOrCreateDocAndToken(docId)) is for the user to provide a doc ID. At the same time, calling this with a `docId` that already exists would result in some unexpected behavior (adding an update to a doc that already has content). I feel like the correct implementation of a `createDocWithContent` function would error if the doc already exists, but I don't think we can do that from the client (and it may not be easy to do it from the server either). So I guess we have two options here:

1. leave as is and don't allow users to provide a `docId` possibly forcing users to keep their own mapping to a doc ID
2. allow users to provide a `docId` but risk them accidentally (and without warning) pushing updates to an existing doc

@michaelsilver @paulgb @feliciachang thoughts?

michaelsilver: I made this explicit design choice based on @paulgb’s input but we can change course if we want

paulgb: Yes, the thinking here was that for now the only way we can guarantee that we are creating a new document is if we generate the doc ID from Y-Sweet. If we allow users to set their own ID, and they set it to a document that already exists, the initial document content will be added into the existing document, which may cause unexpected results (for example, if the initial document is a template for a text doc, it will add the template again to the existing doc)

To alleviate this down the road, I think we should have a server-side endpoint that applies the initial content _only_ if the document is actually new. But we don't have that yet, so we resort to this.

rolyatmax:
Let's also make a ticket to update the y-sweet docs with this new function. (We'll want to add it to [this page](https://docs.jamsocket.com/y-sweet/reference/sdk) and make sure to mention that you should not call `createDocWithContent` within the auth endpoint.)

rolyatmax (approved):

PR #333, opened 2024-11-21
Add create-y-sweet-app

jakelazaroff:
Here's an initial draft of the `create-y-sweet-app` tool. Three major parts:

1. The Next.js template (`js-pkg/create-y-sweet-app/src/templates/nextjs`)
2. The Remix template (`js-pkg/create-y-sweet-app/src/templates/remix`)
3. The command line tool (`js-pkg/create-y-sweet-app/src/run.js`)

Right now the CLI works like this: `npm create y-sweet-app --template nextjs project-name`. It copies the appropriate template to the `project-name` directory and then runs `git init` and `npm install`. That's the only command; anything else will just print help text that says how to use it.

Calling out a couple things:
- ~~I grabbed the arg parsing code from an old project of mine. Very down to switch to something like oclif; I just wanted to get this up and running quickly.~~ Turns out Node has an [arg parser in their standard library](https://nodejs.org/api/util.html#utilparseargsconfig)!
- Some tools prompt the user about initializing a git repository and installing packages (especially the latter since they might not use npm). How much customization do we want to let them do?


js-pkg/create-y-sweet-app/src/templates/nextjs/components/MyCollaborativeApp.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const setPresence = usePresenceSetter();
+  const presence = usePresence();
+
+  const [self, setSelf] = useState(() => ({ color: randomColor() }));
+  useEffect(() => setPresence(self), [setPresence, self]);
+
+  const others = Array.from(presence.entries());
+
+  return (
+    <div
```
feliciachang, line 27: might be helpful for legibility to break this into three componets - hero, presence, todos?

js-pkg/create-y-sweet-app/src/templates/nextjs/.gitignore
rolyatmax, line 1: nit: add a newline to EOF

js-pkg/create-y-sweet-app/src/templates/nextjs/.env.local
rolyatmax, line 1: Do we want to include this in the gitignore?

js-pkg/create-y-sweet-app/src/templates/nextjs/.env.default
```diff
@@ -0,0 +1 @@
+CONNECTION_STRING="y-sweet-connection-string-goes-here"
```
rolyatmax, line 1: Noticing that this env var name differs from the one in `.env.local`. Is that an accident?

jakelazaroff: This was meant to be a template for their "production" `.env` file from before I switched to linking to the Vercel/Remix docs. I'll remove it!

js-pkg/create-y-sweet-app/src/templates/nextjs/.env.local
jakelazaroff: As in, `templates/nextjs/.gitignore`? Yeah I think that makes sense.

js-pkg/create-y-sweet-app/src/templates/nextjs/next-env.d.ts
rolyatmax, line 1: I _think_ these `next-env.d.ts` files are typically gitignored?

js-pkg/create-y-sweet-app/src/templates/remix/.env.local
```diff
@@ -0,0 +1 @@
+Y_SWEET_CONNECTION_STRING="ys://127.0.0.1:8080"
```
rolyatmax, line 1: Just noticing that there's a `.env.default` in the nextjs example but not one here. Is that intentional?

js-pkg/create-y-sweet-app/src/templates/nextjs/public/tip.svg
rolyatmax, line 1: I really like this. Just curious - what's a good way to generate SVGs like this?

js-pkg/create-y-sweet-app/src/templates/remix/app/routes/_index.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function HydrateFallback() {
+  return <div />;
+}
+
+export default function Home() {
+  const docId = "my-doc-id";
+
+  console.log("test");
```
rolyatmax, line 19: Should we drop this?

jakelazaroff: Whoops 😅

js-pkg/create-y-sweet-app/src/templates/nextjs/public/tip.svg
jakelazaroff: Thanks! Not sure what you mean by "generate" — my workflow is usually

1. Draw something in tldraw/Figma/Sketch/etc
2. Run it through [SVGOMG](https://jakearchibald.github.io/svgomg/)
3. Plop it in the app
4. (if necessary) [create a sprite sheet](https://til.jakelazaroff.com/svg/create-an-svg-sprite-sheet/)

rolyatmax: Yes! That's exactly what I was looking for. I didn't even think to see if I could export SVGs from tldraw!

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,63 @@
+import { exec } from 'node:child_process'
+import { cp, readdir } from 'node:fs/promises'
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { parseArgs } from 'node:util'
```
rolyatmax, line 5: 😍

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,63 @@
...
+
+const { values, positionals } = parseArgs({ options: { template: { type: 'string', short: 't' } } })
```
rolyatmax, line 7: It looks like `parseArgs` throws when it encounters unexpected command line flags. Maybe we should wrap in a try/catch so we can show the `help()` output instead of this error output?:

```
node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_PARSE_ARGS_UNKNOWN_OPTION]: Unknown option '--help'
    at new NodeError (node:internal/errors:405:5)
    at checkOptionUsage (node:internal/util/parse_args/parse_args:98:11)
    at node:internal/util/parse_args/parse_args:360:9
    at Array.forEach (<anonymous>)
    at parseArgs (node:internal/util/parse_args/parse_args:357:3)
    at file:///Users/taylor-drifting/code/jamsocket/y-sweet/js-pkg/create-y-sweet-app/src/run.js:7:33
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12) {
  code: 'ERR_PARSE_ARGS_UNKNOWN_OPTION'
}

Node.js v18.20.2
```

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,63 @@
...
+    process.exit(1)
+  }
+
+  // initialize a git repository
+  exec('git init', { cwd: dest })
+
+  // install dependencies
+  exec('npm install', { cwd: dest })
+
+  console.log(`Created y-sweet app in ${name}!`)
```
rolyatmax, line 62: I think we should either pipe the stdout/stderr of `npm install` to the terminal or we should print "Installing npm packages...". (I _think_ I kinda lean towards the former because it shows progress to the user a bit better?)

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,108 @@
...
+  input: process.stdin,
+  output: process.stdout,
+})
+
+/**
+ * @param {string} prompt
+ * @returns {Promise<string>}
+ */
+function question(prompt, defaultValue = '') {
+  return new Promise((resolve) => rl.question(prompt, (result) => resolve(result || defaultValue)))
```
rolyatmax, line 18: The Node stdlib has so many nice features these days!

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,108 @@
...
+    'y',
+  )
+  if (installResponse.toLowerCase() === 'y') install = true
+
+  let git = false
+  const gitResponse = await question(
+    `Do you want to install dependencies with npm? ${gray('(Y/n)')} `,
+    'y',
+  )
+  if (gitResponse.toLowerCase() === 'y') git = true
```
rolyatmax, line 60: Worth adding the `git` and `install` options to the CLI options?

jakelazaroff: I made this a wizard!

rolyatmax: I just meant as a way to skip those questions of the wizard. I think I've seen other wizards do this. Maybe also makes the command more reusable in automated environments? I don't think it's a blocker for release, though!

js-pkg/create-y-sweet-app/src/templates/remix/app/components/Todos.tsx
```diff
@@ -0,0 +1,153 @@
...
+            className="peer"
+            type="checkbox"
+            checked={todos.get("connstring") || false}
+            onChange={(e) => todos.set("connstring", e.target.checked)}
+          />
+          <p className="col-start-2 peer-checked:line-through">
+            Generate a connection string for your new Y-Sweet service.
+          </p>
+          <p className="col-start-2 text-xs text-gray-500">
+            A connection string is like an API key that tells your Next.js app
```
rolyatmax, line 127: replace "Next.js" with "Remix"

rolyatmax:
When running the remix app (and I'm assuming this is the case for the nextjs app, too), I notice that there are `404 /favicon.ico` errors in the server logs. Maybe worth adding a favicon to both of these projects so we avoid showing those?

js-pkg/create-y-sweet-app/src/templates/nextjs/.env.local
```diff
@@ -0,0 +1 @@
+Y_SWEET_CONNECTION_STRING="ys://127.0.0.1:8080"
```
rolyatmax, line 1: Thoughts on changing this to `CONNECTION_STRING` so it matches the instructions the `y-sweet serve` command gives for running with the examples? Might help reduce confusion by being consistent across those.

Screenshot 2024-11-21 at 6 26 58 PM

rolyatmax (commented):
A few comments there, but I'm _really_ liking this so far!

js-pkg/create-y-sweet-app/src/run.js
jakelazaroff: Looks like there's a `strict` option to not throw which still attempts to parse them. `node src/run.js --asdf foo --template nextjs --blah=qwerty` results in this:

```
{
  values: {
    asdf: true,
    template: 'nextjs',
    blah: 'qwerty'
  },
  positionals: [ 'foo' ]
}
```

Think that's okay (silently ignore them) or would you rather just display the help text?

rolyatmax: Ahh - yeah, maybe it's fine to just turn off strict mode.

jakelazaroff: Took your suggestion and bailed out if they enter unknown flags.

js-pkg/create-y-sweet-app/src/templates/remix/app/components/Todos.tsx
jakelazaroff: Whoops, good catch!

js-pkg/create-y-sweet-app/src/shell.js
```diff
@@ -0,0 +1,41 @@
...
+export async function* execute(command, args, options) {
+  const proc = spawn(command, args, options)
+
+  let done = false
+  proc.on('exit', () => {
+    done = true
+  })
+
+  proc.on('error', (err) => {
+    console.log(err)
```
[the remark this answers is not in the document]
jakelazaroff: Not the right suggestion but I'll stop swallowing these errors.

js-pkg/create-y-sweet-app/src/run.js
```diff
@@ -0,0 +1,131 @@
...
+      if (stderr) console.error(stderr)
+    }
+
+    console.log('✅ Installed dependencies!')
+  }
+
+  // initialize a git repository
+  if (git) {
+    const { stop } = spinner('Initializing Git repository...')
+    for await (const { stdout, stderr } of execute('git', ['init', '-q'], { cwd: dest })) {
```
rolyatmax, line 119: Love this `execute` helper

jakelazaroff: async integrators are the bees knees

js-pkg/create-y-sweet-app/src/templates/nextjs/components/Presence.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { usePresence, usePresenceSetter } from "@y-sweet/react";
+import { useEffect, useState } from "react";
+import { randomColor } from "@/lib/colors";
+
+export function Presence() {
+  const setPresence = usePresenceSetter();
+  const presence = usePresence();
+
+  const [self, setSelf] = useState(() => ({ color: randomColor() }));
+  useEffect(() => setPresence(self), [setPresence, self]);
```
rolyatmax, line 13: One thing we should do in the React library is make the `setPresence` function (and others) referentially stable so it doesn't force `useEffect`s to rerun unnecessarily. (I _think_ it's not referentially stable right now.)

js-pkg/create-y-sweet-app/src/run.js
rolyatmax: I have yet to really harness them effectively!

rolyatmax (approved):

PR #335, opened 2024-11-22
Turn off safe mode for next.js demos

paulgb:
Next.js uses safe mode by default, so in development components are rendered multiple times. This leads to a bad experience for users who are poking around:

- The y-sweet debugger console log message is duplicated when running locally
image

- Two websocket connections are established, making it harder to debug.

### DIFF
--- a/examples/nextjs/next.config.js
+++ b/examples/nextjs/next.config.js
@@ -1,6 +1,7 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: 'standalone',
+  reactStrictMode: false,
 }
 
 module.exports = nextConfig


jakelazaroff (approved):
I think this works as a band-aid but medium term we should try to improve `@y-sweet/react` within strict mode, since people who try to use it with a React app in strict mode outside of our demos will still experience the paper cuts.

paulgb:
> I think this works as a band-aid but medium term we should try to improve `@y-sweet/react` within strict mode, since people who try to use it with a React app in strict mode outside of our demos will still experience the paper cuts.

Agreed! Do you know if there's a non-gross way to do that in the context of e.g. ensuring that a log message only goes out once? I guess we could attach some state to `window` but that feels a bit bad.

PR #345, opened 2024-11-28
Migrate Python library to uv

paulgb:
It's fast and seems to be the direction Python projects are going. We also now spin up a server for each test, for better isolation. This required some changes to the piping on the server side, to allow us to use port `0` to get the OS to assign an available port. This means we need to compute the connection string _after_ we have bound the port, so that the connection string contains the actual bound port and not port `0`.

This also sets up Python tests to run in CI/CD, but only when the Python code changes (at least for now).


michaelsilver (changes requested):
I’m a fan of all of it sans migrating to uv. It’s a good tool and seems to be having a moment, but I’m a fan of using boring technologies. I don’t see a need to change build tooling right now. From what I can see, the rest of this PR adds useful functionality, independent of build system

paulgb:
> I’m a fan of all of it sans migrating to uv. It’s a good tool and seems to be having a moment, but I’m a fan of using boring technologies. I don’t see a need to change build tooling right now. From what I can see, the rest of this PR adds useful functionality, independent of build system

I generally agree, but IMHO there are no “boring” package managers for Python, instead there's a mishmash of tools that have never worked well for me. Prior to introducing `uv`, developing on this library was a pain because you had to set up a `virtualenv`, install the dependencies, and remember to activate the virtualenv before running tests. Despite being newer, `uv` already feels like a more mature product than others in the python ecosystem, e.g. I referred to [the docs](https://docs.astral.sh/uv/) while doing this and they are solid.

michaelsilver:
> developing on this library was a pain because you had to set up a virtualenv, install the dependencies, and remember to activate the virtualenv before running tests.

This is how Python development has been done for a decade, I’m honestly confused why suddenly it’s an issue. I’m happy to sit down and help get you setup if you’re running into issues, but here’s a [very quick reference](https://gist.github.com/duonghuuphuc/7939cfbf82d9664274d299fff3d4c205). On the flip side, I see wasted time in migrating and learning a new tool.

paulgb:
> > developing on this library was a pain because you had to set up a virtualenv, install the dependencies, and remember to activate the virtualenv before running tests.
> 
> This is how Python development has been done for a decade, I’m honestly confused why suddenly it’s an issue. I’m happy to sit down and help get you setup if you’re running into issues, but here’s a [very quick reference](https://gist.github.com/duonghuuphuc/7939cfbf82d9664274d299fff3d4c205). On the flip side, I see wasted time in migrating and learning a new tool.
> 
> (EDIT: In case you’re following the guide, note that M-series Macs use /opt/homebrew instead of /usr/local — you can copy my config [here](https://github.com/michaelsilver/dotfiles/blob/master/bashrc#L64))

It's not that it's suddenly an issue; Python dependency management has always been a mess, there's no “boring” track to fall back on. `pyproject.toml` has [only been around](https://peps.python.org/pep-0621/) as a fleshed-out spec for ~4 years, so most of the tooling is still built for a pre-`pyproject.toml` world.

If we are going to be running tests in CI, I want to use lockfiles, which `pip` does not natively support. The official [tool recommendations](https://packaging.python.org/en/latest/guides/tool-recommendations/#lock-files) from the Python packaging authority for lockfiles are `pip-tools` and `pipenv`, both of which are oriented around a pre-`pyproject.toml` world and seem to require a separate source-of-truth for dependencies. Among tools that do respect `pyproject.toml`, `poetry` and `uv` are the main contenders that I'm aware of. Having used both, I am willing to bet on `uv` winning over `poetry`. `uv` may be the newest, but it's had the most advantage of seeing what works and doesn't works in other language ecosystems, and from what I've seen so far it makes the best decisions.

paulgb:
Sounds good, I've fleshed out the readme a bit more to include installing `uv` and give some basics on using it (e.g. that it doesn't need to be used with virtualenv).

> I think the Python ecosystem really screwed itself with the recent move to pyproject.toml, as previously with setup.py there was a straightforward way to manage a requirements.txt file alongside.

I think `pyproject.toml` is a change in the right direction. The prior `setup.py`/`distutils` approach made sense at the time when it pre-dated any sort of package management or virtual environments, but now that we have things like `pip` and virtual environments, it's a pretty archaic way of doing things. In addition to not supporting lockfiles, having the metadata format for a package defined programmatically means that you can't really support things like `uv add` which need to modify the metadata.

michaelsilver (approved):
Everything looks great, thanks for the super helpful README!

PR #346, opened 2024-11-30
Outline Y-Sweet logo

paulgb:
The Y-Sweet logo has a white outline as part of the logo with some gradients that add depth. On white, the outline is completely invisible except for the gradients, which look like flaws. This adds an outline to the logo so that the outline is apparent on a light background. The outline matches the default GitHub "dark mode" color, so it is completely invisible on the default dark mode.

Before:

image

After:

image


paulgb:
Converted to draft until I remove unrelated files.

jakelazaroff (approved):

PR #351, opened 2024-12-03
Allow foreign origins to auth with demo server

jakelazaroff:
This adds CORS headers to the example instance at [demos.y-sweet.dev](https://demos.y-sweet-dev). Right now I'm just allowlisting `demos.y-sweet.dev`, `blocknotejs.org` and `blocknote-main.vercel.app` (their playground website).

- Probably will get rid of `localhost` before merging; I need it to test the BlockNote docs locally but arguably we don't want to let people use this as a free local testing instance.
- Maybe will add Stackblitz? The BlockNote example has a link to open in Stackblitz which won't work unless we allowlist it, but we might not want to for the same reason as `localhost.

### DIFF
--- a/examples/nextjs/src/app/api/auth/route.ts
+++ b/examples/nextjs/src/app/api/auth/route.ts
@@ -2,10 +2,21 @@ import { CONNECTION_STRING } from '@/lib/config'
 import { getOrCreateDocAndToken } from '@y-sweet/sdk'
 import { NextResponse } from 'next/server'
 
+const CORS_HEADERS = {
+  'access-control-allow-origin': '*',
+  'access-control-allow-methods': 'POST',
+  'access-control-allow-headers': 'content-type',
+}
+
+export async function OPTIONS(request: Request) {
+  return new Response(null, { status: 204, headers: CORS_HEADERS })
+}
+
 export async function POST(request: Request) {
   // In a production app, you'd want to validate that the user is authenticated
   // and has access to the given doc.
+
   const { docId } = await request.json()
   const clientToken = await getOrCreateDocAndToken(CONNECTION_STRING, docId)
-  return NextResponse.json(clientToken)
+  return NextResponse.json(clientToken, { headers: CORS_HEADERS })
 }


examples/nextjs/src/app/api/auth/route.ts
```diff
@@ -2,10 +2,42 @@ import { CONNECTION_STRING } from '@/lib/config'
 import { getOrCreateDocAndToken } from '@y-sweet/sdk'
 import { NextResponse } from 'next/server'
 
+const ALLOWED_ORIGINS = new Set([
+  'demos.y-sweet.dev',
+  'blocknotejs.org',
+  'blocknote-main.vercel.app',
+  'localhost',
+])
```
paulgb, line 10: I'm tempted just to drop the hard-coded list of sites, and add a comment saying that in production you'd want to set `access-control-allow-origin` to a specific domain.

jakelazaroff: I'm fine with that, it just means that anyone will be able to use `demos.y-sweet.dev` as their actual live Y-Sweet service.

If you give the go-ahead I'll make the change!

paulgb: Yeah, I'm fine with that, we already provide a generous free tier so they shouldn't have a reason to do that :)

jakelazaroff: I think I'm more worried about people just copying/pasting/deploying to Vercel or something without ever really knowing what's going on. But I suppose those people are probably unlikely to convert anyway, and if a project like that goes viral we can cross that bridge when we get to it.

paulgb: If they deploy this codebase to vercel it will have its own instance of the auth route so it should just hit that, so I think we’re fine, someone would have to intentionally hard-code in the demo server domain to hit the demo server.

examples/nextjs/src/app/api/auth/route.ts
```diff
@@ -2,10 +2,21 @@ import { CONNECTION_STRING } from '@/lib/config'
 import { getOrCreateDocAndToken } from '@y-sweet/sdk'
 import { NextResponse } from 'next/server'
 
+const CORS_HEADERS = {
+  'access-control-allow-origin': '*',
```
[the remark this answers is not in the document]
paulgb: This is a demo, it’s intentional here.

paulgb (approved):

examples/nextjs/src/app/api/auth/route.ts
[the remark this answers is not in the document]
jakelazaroff: yeah get out of here bot

Issue #308, opened 2024-10-08
Subdocument support


paulgb:
Requires:

- Updating how we store documents to allow multiple documents in one `.ysweet` file.
- Implementing a new wire protocol to support subdocuments.

Destreyf:
@paulgb just circling back to this, the project I am working on right now would benefit greatly by this, is this a feature we could possibly sponsor development on?  We're a small company and bootstrapping everything inhouse so it wouldn't be a huge amount, but if you (or the jamsocket team) open to this I would love to see if there's some way to make this work.

Issue #306, opened 2024-10-08
State hooks


paulgb:
Implement react hooks:

- `useSynced()` returns `true` if the state is fully synced with the server
- `useConnected()` returns `true` if the client is connected to the server

paulgb:
Now available in 0.6.4 (#353)

PR #352, opened 2024-12-10
Rewrite provider

paulgb:
This is essentially a complete rewrite of `YSweetProvider`. It incorporates the reconnect logic directly in the provider itself, instead of using the wrapper, which should make it easier to reason about. This paves the way to beginning to extend the protocol (as required for #306 and #308).


js-pkg/client/src/provider.ts
```diff
@@ -210,443 +50,392 @@ type WebSocketPolyfillType = {
...
-  addOnFailureHandler(handler: () => void): () => void {
-    this.onFailureHandlers.push(handler)
-    return () => {
-      this.onFailureHandlers = this.onFailureHandlers.filter((h) => h !== handler)
+  private websocketError(event: Event) {
+    this.emit('connection-error', event)
+  }
+
+  protected emit(eventName: YSweetEvent, data: any = null): void {
+    const listeners = this.listeners.get(eventName)
```
jakelazaroff, line 359: not a huge issue but we can avoid a level of nesting by falling back to an empty set
```suggestion
    const listeners = this.listeners.get(eventName) || new Set()
```

js-pkg/client/src/provider.ts
```diff
@@ -210,443 +50,392 @@ type WebSocketPolyfillType = {
...
-): YSweetProviderParams {
-  // if the clientToken has a token, add it to the provider params
-  const connectParams = providerParams.params ? { ...providerParams.params } : {}
-  if (clientToken.token) connectParams.token = clientToken.token
-  return { ...providerParams, params: connectParams }
-}
+  private setStatus(status: YSweetStatus) {
+    if (status.status === STATUS_DISCONNECTED && this.shouldConnect) {
+      this.connect()
+    }
```
jakelazaroff, line 377: This feels like it would be more appropriate in `websocketClose` (bonus, we can ditch the `status.status === STATUS_DISCONNECTED` check)

js-pkg/client/src/provider.ts
```diff
@@ -210,443 +50,392 @@ type WebSocketPolyfillType = {
...
+      // Ref: https://github.com/yjs/y-websocket/blob/aa4220407bda51ab6282d1291de6493c136c2089/src/y-websocket.js#L404
+      type = EVENT_SYNC
+    }
+
+    if (!this.listeners.has(type)) {
+      this.listeners.set(type, new Set())
+    }
+    this.listeners.get(type)?.add(listener)
+    if (once) {
+      this.once(type, listener)
```
jakelazaroff, line 423: `this.once` calls `this._on(type, listener, true)` so I think this will recurse infinitely.

jakelazaroff (changes requested):

jakelazaroff:
(the nits aren't important but I think the `this.once` and constructor connect logic are both bugs)

js-pkg/client/src/provider.ts
```diff
@@ -1,207 +1,47 @@
...
+import * as Y from 'yjs'
 
-messageHandlers[messageAuth] = (_encoder, decoder, provider, _emitSynced, _messageType) => {
-  authProtocol.readAuthMessage(decoder, provider.doc, (_ydoc, reason) =>
-    permissionDeniedHandler(provider, reason),
-  )
-}
+const messageSync = 0
+const messageQueryAwareness = 3
+const messageAwareness = 1
```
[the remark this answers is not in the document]
paulgb: Yes, good point! I borrowed these from [`y-websocket`](https://github.com/yjs/y-websocket/blob/aa4220407bda51ab6282d1291de6493c136c2089/src/y-websocket.js#L20-L23) and forgot they were in the global scope.

js-pkg/client/src/provider.ts
paulgb: Opted not to, I think this makes sense if we are getting a set to read, but since we are getting a set to write I think it actually obfuscates the purpose of the code when we create a value just to do a throwaway mutation to it.

jakelazaroff: Isn't this a read? We're just iterating through the members and calling them. Regardless nbd, feel free to leave as is

jakelazaroff (approved):
LGTM 🤘🏻

js-pkg/client/src/provider.ts
jakelazaroff: Isn't it a read? We're just iterating through the members and calling them. Regardless nbd, feel free to leave as is

paulgb: ooh sorry I did a ctrl+f in the source and ended up in the wrong place. Yep, that's just a read, I like your approach then.

PR #353, opened 2024-12-11
Implement events for sync status

paulgb:
`y-websocket` has a `synced` flag, but it is a bit misleading because it only tells whether the initial handshake was completed. Knowing whether a document was synced _in general_ for `y-protocols` is a bit tricky, because we would need to compare the state vector and delete set.

When using WebSockets, we can take advantage of the ordering guarantee and synchronous message processing in Y-Sweet to implement a very simple check of sync.

The client stores two numbers, `lastSyncSent` and `lastSyncAcked`. When it sends an update to the server, it also increments `lastSyncSent` and then sends a separate `messageSyncStatus` message containing that number as the payload.

When the server receives a `messageSyncStatus`, it simply echoes it verbatim. When the client receives the `messageSyncStatus` in return, it updates `lastSyncAcked` to the payload of the message.

If `lastSyncAcked` = `lastSyncSent`, we know that all messages since the last update have been processed. Otherwise, there are outstanding changes.

This builds on #352 and will enable #306.

Demo:

https://github.com/user-attachments/assets/60d4aec2-f025-4e84-a594-28fbc84c67cb

### DIFF
--- a/crates/y-sweet/src/server.rs
+++ b/crates/y-sweet/src/server.rs
@@ -17,7 +17,6 @@ use futures::{SinkExt, StreamExt};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use std::{
-    net::SocketAddr,
     sync::{Arc, RwLock},
     time::Duration,
 };
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -42,8 +42,10 @@ const DELAY_MS_BEFORE_RETRY_TOKEN_REFRESH = 3000
  * Note: this should always be a superset of y-websocket's valid statuses.
  * Ref: https://github.com/yjs/y-websocket/blob/aa4220407bda51ab6282d1291de6493c136c2089/README.md?plain=1#L121
  */
-type YSweetStatus = {
-  status: typeof STATUS_CONNECTED | typeof STATUS_DISCONNECTED | typeof STATUS_CONNECTING
+type YSweetStatus = typeof STATUS_CONNECTED | typeof STATUS_DISCONNECTED | typeof STATUS_CONNECTING
+
+type YSweetStatusObject = {
+  status: YSweetStatus
 }
 
 type WebSocketPolyfillType = {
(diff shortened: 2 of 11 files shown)


js-pkg/client/src/provider.ts
```diff
@@ -146,9 +166,30 @@ export class YSweetProvider {
...
+    const encoder = encoding.createEncoder()
+    encoding.writeVarUint(encoder, messageSyncStatus)
+
+    const versionEncoder = encoding.createEncoder()
+    encoding.writeVarUint(versionEncoder, this.lastSyncSent)
+
+    encoding.writeVarUint8Array(encoder, encoding.toUint8Array(versionEncoder))
+    this.websocket.send(encoding.toUint8Array(encoder))
+
+    this.updateSyncedState()
```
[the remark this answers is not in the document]
paulgb: this is intentional, we want to reflect that the state is not synced yet until the server responds

js-pkg/client/src/provider.ts
```diff
@@ -163,56 +208,56 @@ export class YSweetProvider {
...
+        } else if (event === STATUS_ERROR) {
+          this.off(EVENT_CONNECTION_STATUS, statusListener)
+          resolve(false)
+        }
+      }
+
+      this.on(EVENT_CONNECTION_STATUS, statusListener)
+    })
+
+    this.setupWs(clientToken)
```
jakelazaroff, line 235: Would you be opposed to inlining this? All the small methods that are only used once make the connection flow a bit difficult for me to follow.

crates/y-sweet/src/server.rs
```diff
@@ -17,7 +17,6 @@ use futures::{SinkExt, StreamExt};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use std::{
-    net::SocketAddr,
```
paulgb, line 20: unrelated change that snuck in, but it's just removing an unnecessary import

js-pkg/client/src/provider.ts
```diff
@@ -357,35 +405,19 @@ export class YSweetProvider {
...
+    this.setStatus(STATUS_ERROR)
 
-    if (this.shouldConnect && !this.isConnecting) {
-      this.connect()
-    }
+    this.connect()
   }
 
-  protected emit(eventName: YSweetEvent, data: any = null): void {
+  public emit(eventName: YSweetEvent | YWebsocketEvent, data: any = null): void {
```
jakelazaroff, line 420: Why make this public?

jakelazaroff (approved):
Only a couple minor comments but LGTM

js-pkg/client/src/provider.ts
paulgb: It's needed so that `WebSocketCompatLayer` can emit its own events, unfortunately.

PR #360, opened 2024-12-13
More responsive to network outages

paulgb:
When the browser loses connection abruptly (e.g. wifi disconnect), open connections are not immediately closed. I have observed the timeout taking anywhere from 10s (Firefox) to 1:35 (Chrome). (When using a VPN)

The approach I took is to passively monitor incoming messages. As long as we are receiving messages within a certain time threshold, we can assume that we are online. If a time threshold passes without receiving any messages, we _may_ still be connected and it just happens to be a quiet time, so we send a heartbeat message to the server. Since now we _know_ that we should get a message within a reasonable time, we also set a timeout that will close the connection if we don't hear back in time.

### DIFF
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -18,7 +18,16 @@ const MESSAGE_SYNC_STATUS = 102
 
 const RETRIES_BEFORE_TOKEN_REFRESH = 3
 const DELAY_MS_BEFORE_RECONNECT = 500
-const DELAY_MS_BEFORE_RETRY_TOKEN_REFRESH = 3000
+const DELAY_MS_BEFORE_RETRY_TOKEN_REFRESH = 3_000
+
+/** Amount of time without receiving any message that we should send a MESSAGE_SYNC_STATUS message. */
+const MAX_TIMEOUT_BETWEEN_HEARTBEATS = 2_000
+
+/**
+ * Amount of time after sending a MESSAGE_SYNC_STATUS message that we should close the connection
+ * unless any message has been received.
+ **/
+const MAX_TIMEOUT_WITHOUT_RECEIVING_HEARTBEAT = 3_000
 
 // Note: These should not conflict with y-websocket's events, defined in `ws-status.ts`.
 export const EVENT_LOCAL_CHANGES = 'local-changes'
@@ -126,6 +135,9 @@ export class YSweetProvider {
   /** Whether a (re)connect loop is currently running. This acts as a lock to prevent two concurrent connect loops. */
   private isConnecting: boolean = false
 
+  private heartbeatHandle: ReturnType<typeof setTimeout> | null = null
+  private connectionTimeoutHandle: ReturnType<typeof setTimeout> | null = null
+
   constructor(
     private authEndpoint: AuthEndpoint,
     private docId: string,
@@ -148,6 +160,44 @@ export class YSweetProvider {
     if (extraOptions.connect !== false) {
       this.connect()
     }
+  }
+
+  private clearHeartbeat() {
+    if (this.heartbeatHandle) {
+      clearTimeout(this.heartbeatHandle)
+      this.heartbeatHandle = null
+    }
+  }
+
+  private setHeartbeat() {
+    if (this.heartbeatHandle) {
+      return
+    }
+    this.heartbeatHandle = setTimeout(() => {
+      this.checkSync(false)
+      this.setConnectionTimeout()
+      this.heartbeatHandle = null
+    }, MAX_TIMEOUT_BETWEEN_HEARTBEATS)
+  }
+
+  private clearConnectionTimeout() {
+    if (this.connectionTimeoutHandle) {
+      clearTimeout(this.connectionTimeoutHandle)
+      this.connectionTimeoutHandle = null
+    }
+  }
+
+  private setConnectionTimeout() {
+    if (this.connectionTimeoutHandle) {
+      return
+    }
+    this.connectionTimeoutHandle = setTimeout(() => {
+      if (this.websocket) {
+        this.websocket.close()
+        this.setStatus(STATUS_ERROR)
+      }
+      this.connectionTimeoutHandle = null
+    }, MAX_TIMEOUT_WITHOUT_RECEIVING_HEARTBEAT)
   }
 
   private send(message: Uint8Array) {
@@ -182,12 +232,14 @@ export class YSweetProvider {
       syncProtocol.writeUpdate(encoder, update)
       this.send(encoding.toUint8Array(encoder))
 
-      this.checkSync()
+      this.checkSync(true)
     }
   }
 
-  private checkSync() {
-    this.lastSyncSent += 1
+  private checkSync(increment: boolean) {
+    if (increment) {
+      this.lastSyncSent += 1
+    }
     const encoder = encoding.createEncoder()
     encoding.writeVarUint(encoder, MESSAGE_SYNC_STATUS)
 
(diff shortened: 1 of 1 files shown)


js-pkg/client/src/provider.ts
```diff
@@ -143,21 +154,83 @@ export class YSweetProvider {
...
+  private setHeartbeat() {
+    if (this.heartbeatHandle) {
+      return
+    }
+    this.heartbeatHandle = setTimeout(() => {
+      this.checkSync()
+      this.setConnectionTimeout()
+      this.heartbeatHandle = null
+    }, MAX_TIMEOUT_BETWEEN_HEARTBEATS)
+  }
```
jakelazaroff, line 203: Trying to think if there's ever a scenario in which we want to bail out of setting a heartbeat if there's one queued, rather than just resetting it? I think we basically always want a heartbeat going and to just keep "moving the timer up" when we receive messages from the server.

I'd change these methods like this:

```suggestion
  private clearHeartbeat() {
    if (this.heartbeatHandle) {
      clearTimeout(this.heartbeatHandle)
      this.heartbeatHandle = null
    }
  }

  private resetHeartbeat() {
    this.clearHeartbeat()

    this.heartbeatHandle = setTimeout(() => {
      this.checkSync()
      this.setConnectionTimeout()
      this.heartbeatHandle = null
    }, MAX_TIMEOUT_BETWEEN_HEARTBEATS)
  }
```

And then we should probably call `clearHeartbeat` in `websocketClose` and `websocketError`.

jakelazaroff (commented):
To sum up the comments, I think we can simplify this a little bit by:

1. Rather than being able to "set" and "clear" a heartbeat, just have one operation — "reset", which clears any current timeout and sets another one. That gets called on websocket connection and on receiving a message. (We might still want a "clear" that gets called in situations where we don't want to set another one, like _actually_ going offline). 
2. Rather than just looking for a connection timeout on specific sync messages, set it on _every_ sent message. This means we'll initiate the reconnection flow if it takes too long to get a response to any message. (This isn't technically true: if we send two messages and then get one response, it will clear the connection timeout without setting another, but we'd fall back to the heartbeat anyway in that scenario).

Also, I can't comment on lines that haven't changed, but I think we should clear both the heartbeat and connection timeout in `websocketClose` and `websocketError`

js-pkg/client/src/provider.ts
```diff
@@ -143,21 +154,83 @@ export class YSweetProvider {
...
   }
 
+  private offline() {
+    // When the browser indicates that we are offline, we immediately
+    // probe the connection status.
+    // This accelerates the process of discovering we are offline, but
+    // doesn't mean we entirely true the browser, since it can be wrong
+    // (e.g. in the case that the connection is over localhost).
+    this.checkSync()
+    this.setConnectionTimeout()
```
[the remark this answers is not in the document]
paulgb: The trouble with that is that not every message necessarily has a response, so we might encounter false positives.

That said, I have moved the call into `checkSync`, since we know that every `checkSync` call should return a response. This also should speed up offline detection if a user is interacting with the state when the connection is dropped.

js-pkg/client/src/provider.ts
```diff
@@ -143,21 +154,83 @@ export class YSweetProvider {
...
       this.websocket.send(message)
     }
   }
 
   private updateSyncedState() {
-    let hasLocalChanges = this.lastSyncAcked !== this.lastSyncSent
+    let hasLocalChanges = this.ackedVersion !== this.localVersion
     if (hasLocalChanges === this.hasLocalChanges) {
       return
     }
```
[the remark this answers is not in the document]
paulgb: I like that, done. Unfortunately it adds a bit of complexity at the former call sites because we want to emit if and only if the `hasLocalChanges` value would change, but I do think it's an improvement overall.

paulgb:
> To sum up the comments, I think we can simplify this a little bit by:
> 
> 1. Rather than being able to "set" and "clear" a heartbeat, just have one operation — "reset", which clears any heartbeat timeout and sets another one. That gets called on websocket connection and on receiving a message. (We might still want a "clear" that gets called in situations where we don't want to set another one, like _actually_ going offline).
> 2. Rather than just looking for a connection timeout on specific sync messages, set it on _every_ sent message. This means we'll initiate the reconnection flow if it takes too long to get a response to any message. (This isn't technically true: if we send two messages and then get one response, it will clear the connection timeout without setting another, but we'd fall back to the heartbeat anyway in that scenario).
> 
> Also, I can't comment on lines that haven't changed, but I think we should clear both the heartbeat and connection timeout in `websocketClose` and `websocketError`

Thanks for the review! I think I did all of this, except for one thing: instead of setting the connection timeout only on every message type, I do it on every _sync_ message. This way if we send some message that the server doesn't need to acknowledge, we don't accidentally disconnect. This change still cleans up the logic a bit.

js-pkg/client/src/provider.ts
```diff
@@ -143,27 +151,101 @@ export class YSweetProvider {
...
+      clearTimeout(this.heartbeatHandle)
+      this.heartbeatHandle = null
+    }
+  }
+
+  private resetHeartbeat() {
+    this.clearHeartbeat()
+    this.heartbeatHandle = setTimeout(() => {
+      this.checkSync()
+      this.setConnectionTimeout()
```
jakelazaroff, line 194: This should no longer be necessary since we now set the timeout in `this.checkSync`
```suggestion
```

js-pkg/client/src/provider.ts
```diff
@@ -182,22 +264,21 @@ export class YSweetProvider {
...
     encoding.writeVarUint(encoder, MESSAGE_SYNC_STATUS)
 
     const versionEncoder = encoding.createEncoder()
-    encoding.writeVarUint(versionEncoder, this.lastSyncSent)
+    encoding.writeVarUint(versionEncoder, this.localVersion)
     encoding.writeVarUint8Array(encoder, encoding.toUint8Array(versionEncoder))
 
     this.send(encoding.toUint8Array(encoder))
-
-    this.updateSyncedState()
```
jakelazaroff, line 200: Just making sure this is intentional — we do _not_ want to emit the `local-changes` event here?

js-pkg/client/src/provider.ts
```diff
@@ -398,6 +485,7 @@ export class YSweetProvider {
   private websocketClose(event: CloseEvent) {
     this.emit(EVENT_CONNECTION_CLOSE, event)
     this.setStatus(STATUS_ERROR)
+    this.clearHeartbeat()
```
jakelazaroff, line 488: I think we want to clear the connection timeout here as well:
```suggestion
    this.clearHeartbeat()
    this.clearConnectionTimeout()
```

js-pkg/client/src/provider.ts
```diff
@@ -413,6 +501,7 @@ export class YSweetProvider {
   private websocketError(event: Event) {
     this.emit(EVENT_CONNECTION_ERROR, event)
     this.setStatus(STATUS_ERROR)
+    this.clearHeartbeat()
```
jakelazaroff, line 504: Ditto re: clearing connection timeout
```suggestion
    this.clearHeartbeat()
    this.clearConnectionTimeout()
```

js-pkg/client/src/provider.ts
paulgb: that's right, this function is now responsible for checking if the _current_ `localVersion` has been seen by the server. Since it is no longer responsible for incrementing `localVersion`, it can no longer lead to the local sync status changing -- it is the responsibility of the caller to increment `localVersion` and emit the event if necessary.

jakelazaroff (approved):
LGTM 🤘🏻

PR #361, opened 2024-12-13
Prepare client package for jsr.io

paulgb:
Seems worth trying out, in part for its built-in documentation viewer. If it works well, I'm tempted to remove docs.y-sweet.dev in favor of linking to jsr.io docs.


paulgb:
Looks like it won't work until jsr supports peer dependencies (https://github.com/jsr-io/jsr/issues/301)

PR #362, opened 2024-12-17
small refactor to clarify logic

rolyatmax:
Small logic clarification I came across as I was catching up on the new provider. `getClientToken` already handles the `AuthEndpoint` type for both cases: when it's a `function` and when it's a `string`.


michaelsilver (approved):

jakelazaroff (approved):
LGTM

paulgb (approved):
Thanks!

PR #363, opened 2024-12-18
Ensure new provider works with older server versions

paulgb:
The new client relies on a message type in order to support heartbeats and sync status. If it's talking to an older server version, the server will ignore its requests for heartbeats, and the client will assume the connection is unhealthy and attempt to reconnect in a periodic loop.

To test this, run the examples along with an old version of the server, e.g. `npx y-sweet@0.6.0 serve`

This fixes the library by adding a boolean field `receivedAtLeastOneSyncResponse`. As long as this is `false`, we will continue to send heartbeats and sync status requests, but we will never terminate a connection over a missed heartbeat.

This is a somewhat crude approach. We should support version negotiation eventually. But it's nice and simple, and will be easy to rip out later.

### DIFF
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -148,6 +148,17 @@ export class YSweetProvider {
 
   private retries: number = 0
 
+  /**
+   * Older versions of the Y-Sweet server did not support the sync message, and would ignore it.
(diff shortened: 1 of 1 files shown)


jakelazaroff (approved):
LGTM

rolyatmax (approved):

PR #369, opened 2025-01-06
Fix debugger console logging in React hook

paulgb:
The new Y-Sweet provider broke the code that printed the Y-Sweet debugger to the console, because the code that reads the active client token runs before the client token is fetched (since fetching now happens asynchronously). This fixes the debugger by installing an event listener on the provider, which is triggered when the client token becomes available.

Reported by Sorin [on discord](https://discord.com/channels/939641163265232947/1202346195142312019/1325396209770303489).

### DIFF
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -1,4 +1,4 @@
-import { ClientToken } from '@y-sweet/sdk'
+import { encodeClientToken, type ClientToken } from '@y-sweet/sdk'
 import * as decoding from 'lib0/decoding'
 import * as encoding from 'lib0/encoding'
 import * as awarenessProtocol from 'y-protocols/awareness'
@@ -91,6 +91,9 @@ export type YSweetProviderParams = {
    * Defaults to `false`; set to `true` to enable.
    */
   offlineSupport?: boolean
+
+  /** Whether to show the debugger link. Defaults to true. */
+  showDebuggerLink?: boolean
 }
 
 async function getClientToken(authEndpoint: AuthEndpoint, roomname: string): Promise<ClientToken> {
@@ -143,6 +146,7 @@ export class YSweetProvider {
   private connectionTimeoutHandle: ReturnType<typeof setTimeout> | null = null
 
   private reconnectSleeper: Sleeper | null = null
+  private showDebuggerLink = true
 
   private indexedDBProvider: IndexedDBProvider | null = null
 
@@ -159,6 +163,13 @@ export class YSweetProvider {
    */
   private receivedAtLeastOneSyncResponse: boolean = false
 
+  private get debugUrl() {
+    if (!this.clientToken) return null
+
+    const payload = encodeClientToken(this.clientToken)
+    return `https://debugger.y-sweet.dev/?payload=${payload}`
+  }
+
   constructor(
     private authEndpoint: AuthEndpoint,
     private docId: string,
@@ -168,6 +179,8 @@ export class YSweetProvider {
     if (extraOptions.initialClientToken) {
       this.clientToken = extraOptions.initialClientToken
     }
+
+    this.showDebuggerLink = extraOptions.showDebuggerLink !== false
 
     // Sets up some event handlers for y-websocket compatibility.
     new WebSocketCompatLayer(this)
@@ -371,7 +384,9 @@ export class YSweetProvider {
     this.isConnecting = true
     this.setStatus(STATUS_CONNECTING)
 
-    while (![STATUS_OFFLINE, STATUS_CONNECTED].includes(this.status)) {
+    const lastDebugUrl = this.debugUrl
+
+    connecting: while (![STATUS_OFFLINE, STATUS_CONNECTED].includes(this.status)) {
       this.setStatus(STATUS_CONNECTING)
       let clientToken
       try {
@@ -391,7 +406,7 @@ export class YSweetProvider {
       for (let i = 0; i < RETRIES_BEFORE_TOKEN_REFRESH; i++) {
         if (await this.attemptToConnect(clientToken)) {
           this.retries = 0
-          break
+          break connecting
         }
 
         let timeout =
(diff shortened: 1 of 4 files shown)


js-pkg/react/src/main.tsx
```diff
@@ -253,26 +259,40 @@ export function YDocProvider(props: YDocProviderProps) {
...
+            'font-style: italic;',
+          )
+        }
       }
 
+      provider.on('connection-status', connectionCallback)
+
       if (canceled) {
+        provider.off('connection-status', connectionCallback)
         provider.destroy()
```
rolyatmax, line 296: I think the function returned from the `useEffect()` should handle all the cleanup (i.e. destroying the provider). Ideally, destroying the provider would also make sure all event subscriptions are removed. The original purpose of the `canceled` check here, I think, is to drop out of the execution following an `await` in case the promise chain had been canceled in the meantime.

In other words, I think we should (1) make sure `provider.destroy()` removes all event subscriptions, (2) remove the logic from this `if (canceled)` block except the `return`.

js-pkg/react/src/main.tsx
```diff
@@ -253,26 +259,40 @@ export function YDocProvider(props: YDocProviderProps) {
...
-        console.log(
-          '%cTo hide the debugger link, pass showDebuggerLink={false} to YDocProvider',
-          'font-style: italic;',
-        )
+      // If the client reconnects, the "connected" event will fire again. We should only show the
+      // debugger link again if the client token has changed.
+      let lastClientToken: { value: ClientToken | null } = { value: null }
+
+      let connectionCallback = (status: YSweetStatus) => {
+        if (props.showDebuggerLink === false) return // note: if showDebuggerLink is undefined, we show the debugger link.
```
rolyatmax, line 273: There is potential for a bug here as this closes over the old `props` value. If that `props.showDebuggerLink` value changes, this closure will still use the old value.

Here's an idea: do we only want to show the debugger link for React users? Maybe we should handle this inside the Provider (in `@y-sweet/client`) instead of in React. I think it might be easier to reason about the state there.

jakelazaroff: +1 to showing the link in `@y-sweet/client`.

Not strictly related to this PR but imo one thing that makes this a little confusing is the IIFE (which isn't necessary since `createYjsProvider` isn't an async function)

js-pkg/react/src/main.tsx
```diff
@@ -241,6 +242,11 @@ export type YDocProviderProps = {
   offlineSupport?: boolean
 }
 
+function clientTokensEqual(a: ClientToken | null, b: ClientToken | null): boolean {
+  if (a === null || b === null) return false
```
jakelazaroff, line 246: Feels kind of weird to allow `null` here — like, if _both_ `a` and `b` are `null`, isn't it kinda… vacuously true, in the same sense that `[].every(anyFunction)` is? I'd push the `null` checking up to the caller.

js-pkg/react/src/main.tsx
rolyatmax: Oh yeah! Totally forgot that we made `createYjsProvider` synchronous!

jakelazaroff:
@rolyatmax @paulgb Updated this PR to move the debugger URL logic to `@y-sweet/client`. In general this feels a lot better to me — great suggestion Taylor!

Note that I'm making the `debugUrl` getter private, since our plan is to eventually deprecate it and provide embedded dev tools.

paulgb:
> @rolyatmax @paulgb Updated this PR to move the debugger URL logic to `@y-sweet/client`. In general this feels a lot better to me — great suggestion Taylor!
> 
> Note that I'm making the `debugUrl` getter private, since our plan is to eventually deprecate it and provide embedded dev tools.

How about making it public but marking with jsdoc as `@deprecated`, so that people aren't surprised if it goes away but it can solve some problems in the meantime?

js-pkg/client/src/provider.ts
```diff
@@ -407,6 +422,17 @@ export class YSweetProvider {
...
 
     this.isConnecting = false
+
+    if (this.showDebuggerLink && lastDebugUrl !== this.debugUrl) {
+      console.log(
+        `%cOpen this in Y-Sweet Debugger ⮕ ${this.debugUrl}`,
+        'font-size: 1.5em; display: block; padding: 10px;',
+      )
+      console.log(
+        '%cTo hide the debugger link, pass showDebuggerLink={false} to YDocProvider',
```
paulgb, line 432: Now that this is in the client we can't assume the user is using react, right? Maybe we need to cover both cases (react and native) here?

jakelazaroff: good catch, I'll just make the message a bit more generic

js-pkg/react/src/main.tsx
```diff
@@ -250,38 +241,15 @@ export function YDocProvider(props: YDocProviderProps) {
   const [ctx, setCtx] = useState<YjsContextType | null>(null)
 
   useEffect(() => {
-    let canceled = false
-    let provider: YSweetProvider | null = null
     const doc = new Y.Doc()
+    const provider = createYjsProvider(doc, docId, authEndpoint, {
+      initialClientToken,
+      offlineSupport: props.offlineSupport,
+    })
```
rolyatmax, line 248: Do we need to pass in `showDebuggerLink` here now?

js-pkg/react/src/main.tsx
```diff
@@ -250,38 +241,15 @@ export function YDocProvider(props: YDocProviderProps) {
...
-        return
-      }
-
-      setCtx({ doc, provider })
-    })()
+    setCtx({ doc, provider })
 
     return () => {
-      canceled = true
       provider?.destroy()
```
rolyatmax, line 253: Nit: now that all the logic here is synchronous, we can drop the `?` after `provider`.

rolyatmax (approved):
Two small comments there, but otherwise LGTM! Thanks for taking this over the finish line!

js-pkg/react/src/main.tsx
jakelazaroff: whoops good catch

PR #373, opened 2025-01-15
implement read-only auth

rolyatmax:
Have left a few comments on this PR to give more context. Probably easiest to review this PR commit-by-commit, leveraging the context given in each commit message.

### DIFF
--- a/crates/y-sweet-core/src/api_types.rs
+++ b/crates/y-sweet-core/src/api_types.rs
@@ -1,7 +1,4 @@
-use std::collections::HashMap;
-
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
 
 #[derive(Serialize)]
 pub struct NewDocResponse {
@@ -27,13 +24,12 @@ impl Authorization {
 
 #[derive(Deserialize)]
 #[allow(unused)]
+#[serde(default)]
 pub struct AuthDocRequest {
     #[serde(default = "Authorization::full")]
     pub authorization: Authorization,
     #[serde(rename = "userId")]
     pub user_id: Option<String>,
-    #[serde(default)]
-    pub metadata: HashMap<String, Value>,
     #[serde(skip_serializing_if = "Option::is_none", rename = "validForSeconds")]
     pub valid_for_seconds: Option<u64>,
 }
@@ -43,7 +39,6 @@ impl Default for AuthDocRequest {
         Self {
             authorization: Authorization::Full,
             user_id: None,
-            metadata: HashMap::new(),
             valid_for_seconds: None,
         }
     }
--- a/crates/y-sweet-core/src/api_types.rs
+++ b/crates/y-sweet-core/src/api_types.rs
@@ -6,7 +6,7 @@ pub struct NewDocResponse {
     pub doc_id: String,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Copy, Clone, Serialize, Deserialize)]
 pub enum Authorization {
     #[serde(rename = "read-only")]
     ReadOnly,
@@ -59,6 +59,10 @@ pub struct ClientToken {
     /// An optional token that can be used to authenticate the client to the server.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub token: Option<String>,
+
+    /// The authorization level of the client.
+    #[serde(rename = "authorization")]
+    pub authorization: Authorization,
 }
 
 #[derive(Deserialize, Debug)]
--- a/crates/y-sweet-core/src/api_types.rs
+++ b/crates/y-sweet-core/src/api_types.rs
@@ -8,9 +8,7 @@ pub struct NewDocResponse {
 
 #[derive(Serialize, Deserialize)]
 pub enum Authorization {
-    #[serde(rename = "none")]
-    Nothing,
-    #[serde(rename = "readonly")]
+    #[serde(rename = "read-only")]
     ReadOnly,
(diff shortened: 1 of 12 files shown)


crates/y-sweet-core/src/api_types.rs
```diff
@@ -1,19 +1,14 @@
...
     pub doc_id: String,
 }
 
-#[derive(Deserialize)]
+#[derive(Serialize, Deserialize)]
 pub enum Authorization {
-    #[serde(rename = "none")]
-    Nothing,
-    #[serde(rename = "readonly")]
+    #[serde(rename = "read-only")]
```
rolyatmax, line 11: This matches the types in the JS client library

crates/y-sweet-core/src/api_types.rs
```diff
@@ -27,13 +22,12 @@ impl Authorization {
 
 #[derive(Deserialize)]
 #[allow(unused)]
+#[serde(default)]
```
rolyatmax, line 25: I believe this means if someone passes an unknown field (e.g. the old `metadata` field), deserialization will still succeed.

crates/y-sweet-core/src/api_types.rs
```diff
@@ -27,13 +22,12 @@ impl Authorization {
...
 pub struct AuthDocRequest {
     #[serde(default = "Authorization::full")]
     pub authorization: Authorization,
     #[serde(rename = "userId")]
     pub user_id: Option<String>,
-    #[serde(default)]
-    pub metadata: HashMap<String, Value>,
```
rolyatmax, line 36: We don't use this at present. And if we did, we'd likely need to encode it into the token which we'd like to avoid since its length is unbounded. Let's just drop this field altogether.

crates/y-sweet/src/server.rs
```diff
@@ -440,37 +431,50 @@ async fn get_doc_as_update(
 async fn get_doc_as_update_deprecated(
     Path(doc_id): Path<String>,
     State(server_state): State<Arc<Server>>,
-    authorization: Option<TypedHeader<headers::Authorization<headers::authorization::Bearer>>>,
+    auth_header: Option<TypedHeader<headers::Authorization<headers::authorization::Bearer>>>,
```
rolyatmax, line 434: A lot of these changes are just renaming vars/args to better differentiate between an Authorization (an enum that indicates read/write access) and an auth header.

crates/y-sweet-core/src/doc_connection.rs
```diff
@@ -160,12 +172,24 @@ impl DocConnection {
...
                 SyncMessage::Update(update) => {
-                    let mut awareness = a.write().unwrap();
-                    protocol.handle_update(&mut awareness, Update::decode_v1(&update)?)
+                    if can_write {
+                        let mut awareness = a.write().unwrap();
+                        protocol.handle_update(&mut awareness, Update::decode_v1(&update)?)
+                    } else {
+                        Err(sync::Error::PermissionDenied {
+                            reason: "Token does not have write access".to_string(),
+                        })
```
jakelazaroff, line 191: Do we plan to make the clients aware of the authorization level as well? Not sure how these errors are logged/captured/etc but as of now this is an expected code path, since clients will send updates as though they have full write access.

crates/y-sweet-core/src/api_types.rs
```diff
@@ -66,6 +59,10 @@ pub struct ClientToken {
     /// An optional token that can be used to authenticate the client to the server.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub token: Option<String>,
+
+    /// The authorization level of the client.
+    #[serde(rename = "authorization")]
+    pub authorization: Authorization,
```
rolyatmax, line 65: We're now returning `authorization` as part of the `ClientToken` so that clients can limit front-end functionality (or just render differently) when the token is `read-only`.

crates/y-sweet/src/server.rs
```diff
@@ -385,7 +385,7 @@ impl Server {
             if let Some(token) = token {
                 let authorization = authenticator
                     .verify_doc_token(token, doc, current_time_epoch_millis())
-                    .map_err(|e| (StatusCode::FORBIDDEN, e))?;
+                    .map_err(|e| (StatusCode::UNAUTHORIZED, e))?;
```
paulgb, line 388: is this a case where the test is wrong rather than the implementation? I think if the token is provided which it seems like the guard checks here), but the token is invalid, it's a forbidden rather than unauthorized.

@ellipsis-dev do you know?

rolyatmax: I think if `verify_doc_token` fails, it's because the token is invalid - which should return a 401 - a failure to _authenticate_. (I hate that the spec calls it `401 Unauthorized` since 403s should be used when _authorization_ fails while 401s should be used when _authentication_ fails. 🤷‍♂️)

rolyatmax: From MDN docs:

The HTTP `401 Unauthorized` [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) status code indicates that a request was not successful because it lacks _valid_ authentication credentials for the requested resource.

A `401 Unauthorized` is similar to the [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.

Issue #386, opened 2025-01-28
[@y-sweet/client] `provider.disconnect()` doesn't work


dtkav:
`provider.disconnect()` is broken in the client because the handler for `websocket.onclose` calls `connect()` indiscriminately:
https://github.com/jamsocket/y-sweet/blob/44f4608004b35145d6f8885b87da0752e56f13d7/js-pkg/client/src/provider.ts#L585C1-L585C19

I had to call `(provider as any).websocket.onclose = null` before `disconnect()` in order to get it to work.

I think this is why separating intent (`shouldConnect`) from status is important.
`shouldConnect` is really not the same thing as `status == 'offline'` as is suggested by the deprecation notice.

Suggested fix is to bring back a settable `shouldConnect` property. Set this when the user calls `disconnect()` and then check it in the `websocket.onclose` and `websocket.onerror` handlers. Only reconnect if `shouldConnect` is true.

paulgb:
Thanks for catching this.

The [disconnect function](https://github.com/jamsocket/y-sweet/blob/44f4608004b35145d6f8885b87da0752e56f13d7/js-pkg/client/src/provider.ts#L455-L461) sets the status to `STATUS_OFFLINE`, but it seems that what's happening is that the onclose handler has already been called by the time we get there, so we enter [this loop](https://github.com/jamsocket/y-sweet/blob/44f4608004b35145d6f8885b87da0752e56f13d7/js-pkg/client/src/provider.ts#L406) which we should not.

We could also guard against `connect` being called in the first place from the `onclose` handler if the status is offline, but it wouldn't solve the problem because it would also be called before the state update. I think switching the order they are called in the disconnect function should switch it.

I generally agree about separating state from intent, and if this were any more complex of a state machine I would, because treating intent as state tends to explode the set of possible states, but in this case it's essentially the same set of states -- if `shouldConnect` is false we should always be offline; if `shouldConnect` is true we are either connected or should be trying to be.

elliotmoose:
Hey there @paulgb! Sorry to revive this, but I think the issue is still present. Calling `this.setStatus(STATUS_OFFLINE)` before `this.websocket.close()` does nothing, because we call `this.setStatus(STATUS_CONNECTING)` here before the `while` loop which overrides the intent: https://github.com/jamsocket/y-sweet/blob/4f1909b2c3390ced54d32b82a673f4f7ee8ac780/js-pkg/client/src/provider.ts#L409-L414

PR #388, opened 2025-02-05
Support warning of unsaved changes

paulgb:
This adds a `warnOnClose` parameter to both the `YSweetProvider` and React component. If it is set to `true`, and the user closes the tab, they will see a warning like this:

Screenshot 2025-02-05 at 10 25 39 AM

Also includes a fix for #386

### DIFF
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -186,7 +186,6 @@ export class YSweetProvider {
     private doc: Y.Doc,
     extraOptions: Partial<YSweetProviderParams> = {},
   ) {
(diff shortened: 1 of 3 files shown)


js-pkg/client/src/provider.ts
```diff
@@ -183,6 +186,7 @@ export class YSweetProvider {
     private doc: Y.Doc,
     extraOptions: Partial<YSweetProviderParams> = {},
   ) {
+    this.handleBeforeUnload = this.handleBeforeUnload.bind(this)
```
rolyatmax, line 189: nit: may we should put this down by the other `bind()` statements?

PR #416, opened 2025-09-27
only persist documents to S3 when dirty

lxe:
Fix an issue where y-sweet was persisting documents to S3 every ~10 seconds even when no changes had occurred.

From what I understand, the persistence worker wakes up both on the dirty signal and also on a `checkpoint_freq` interval and was always calling persist, regardless of the dirtiness, which would write it to S3 regardless of whether any changes had been made. 

The fix adds a check to only persist when the document is actually dirty. 

I'm still a little bit unclear behind the reasoning for the logic of signaling the worker both on a dirty callback and on a timer.

So I wonder if we instead just start the worker once, wake every 10 seconds, check if dirty, persist if needed.

### DIFF
--- a/crates/y-sweet-core/src/sync_kv.rs
+++ b/crates/y-sweet-core/src/sync_kv.rs
@@ -56,7 +56,17 @@ impl SyncKv {
         }
     }
 
+    pub fn is_dirty(&self) -> bool {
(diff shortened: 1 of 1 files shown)


rolyatmax (approved):
Thanks for this fix!

PR #417, opened 2025-09-29
use SeqCst ordering for sync_kv dirty flag

rolyatmax:
Now that we're not persisting unless the document is marked dirty, let's use an atomic ordering with stronger guarantees for the KV store's `dirty` flag.


crates/y-sweet-core/src/sync_kv.rs
```diff
@@ -50,14 +50,14 @@ impl SyncKv {
     }
 
     fn mark_dirty(&self) {
-        if !self.dirty.load(Ordering::Relaxed) && !self.shutdown.load(Ordering::SeqCst) {
-            self.dirty.store(true, Ordering::Relaxed);
+        if !self.is_dirty() && !self.shutdown.load(Ordering::SeqCst) {
+            self.dirty.store(true, Ordering::SeqCst);
             (self.dirty_callback)();
         }
```
paulgb, line 56: Good catch but technically there is still a bug here, because the value of `self.dirty` could change between the read and the write. We could use an atomic swap instead, like this:

```rust
if !self.shutdown.load(Ordering::SeqCst) {
    if !self.dirty.swap(true, Ordering::SeqCst) {
        (self.dirty_callback)();
    }
}
```

PR #425, opened 2025-12-30
Fix provider destroy function to properly cleanup connections (#394)

### LINKED ISSUE #394: Provider destroy function broken
After migration from 0.6.3 to 0.8.2, provider.destroy() just temporally changes the connection status, but keeps living, firing unwanted events and causing memory leaks. There's no way right now to destroy a provider.

P.S.: Also, receiving some sort of structured changelog would be kindly appreciated

fabiorush:
Previously, destroy() would temporarily change status but the provider would continue living, firing unwanted events and causing memory leaks. The websocketClose handler would trigger reconnection even during graceful shutdown.

This fix ensures proper cleanup by setting status to OFFLINE before closing the websocket, clearing timers, and waking pending reconnection attempts. Also adds a guard in websocketClose to only reconnect if status is not OFFLINE.

Fixes #394

### DIFF
--- a/js-pkg/client/src/provider.ts
+++ b/js-pkg/client/src/provider.ts
@@ -588,10 +588,12 @@ export class YSweetProvider {
 
   private websocketClose(event: CloseEvent) {
     this.emit(EVENT_CONNECTION_CLOSE, event)
-    this.setStatus(STATUS_ERROR)
-    this.clearHeartbeat()
-    this.clearConnectionTimeout()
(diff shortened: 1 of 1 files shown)


fabiorush:
Hi @paulgb, this PR is ready for review whenever you have a moment!

fabiorush:
Hi @rolyatmax, this PR is ready for review whenever you have a moment!

